PR #37, opened 2026-06-11
Add optional tamper-evident audit log for av

lofimichael:
## Summary

Adds an **optional, tamper-evident audit log** for the `av` CLI, closing the gap where **secret pulls were never recorded** and only `av contain` had a durable (GUI-daemon) trail. Default **OFF** (opt-in).

Covers:
- **Secret pulls** — `inject`, `credential-helper`, `dotenv run|export` (pre-exec / pre-spawn).
- **Command invocations** — `gate` (previously had *no* durable record) and `contain`.

## How it works

- **Format:** newline-delimited JSON (JSONL) at `~/Library/Logs/av/audit.jsonl` (dir `0700`, file `0600`). `jq`/`grep`/`tail`-friendly.
- **Exec-safe & non-blocking:** records are written synchronously *before* the `execve`/`sandbox.exec()`/spawn that replaces or detaches the process (no background thread can survive that). A single `O_APPEND` write, a **non-blocking** `flock(LOCK_EX|LOCK_NB)` on a dedicated lock file (a contended writer drops its own record rather than stalling the gated op), no `fsync`, and **all errors swallowed** — auditing can never delay or break a gated operation.
- **Tamper-evident:** always-on SHA-256 hash chain + monotonic seq; optional Keychain-keyed HMAC layer (`AV_AUDIT_HMAC=1`) for forgery-resistance against anyone without the key. The HMAC key is provisioned out-of-band via `av audit setup`, so the exec path only *reads* it (never prompts mid-injection).
- **Structural redaction:** the record body type has **no field that can hold a secret value** — only key *names*, paths, argv (droppable via `AV_AUDIT_NO_ARGV`, since a user can put a secret on the command line), digests, fingerprints, decisions. The credential-helper token is recorded as a presence bool, never the value.
- **Zero new dependencies** (reuses `serde`/`serde_json`/`sha2`/`libc`; hand-rolled HMAC + RFC3339).

## Toggling

Persistent, CLI-only:

```sh
av audit enable      # writes ~/Library/Application Support/Automic Vault/audit-config.json (0600)
av audit disable
av log               # view recent records (-n N, --event KIND, --json, --verify)
av audit verify      # check the hash chain (+ HMAC if enabled); non-zero exit on a break
av audit path        # show path, config, enabled/HMAC state
av audit setup       # provision the optional Keychain HMAC key (needs AV_AUDIT_HMAC=1)
```

**Resolution order:** `AV_AUDIT` env (`1`/`0`, one-shot/CI) > persisted config file > **default OFF**.

## Security model (honest)

Tamper-**evident**, not tamper-**proof** against the file's own (same-uid) owner: they can delete the file (undetectable from the file alone), and the keyless chain can be rewritten by a local editor — the HMAC layer closes that only against attackers without the Keychain key. No `fsync` means a crash can drop the last record(s). Real WORM/forensic guarantees would need off-host shipping or `chflags`; explicitly out of scope for v1.

## Testing

13 unit tests: SHA-256 & HMAC (RFC 4231) vectors, tamper + HMAC-forgery detection, on-disk round-trip, rotation-spans-chain, **non-blocking contention**, config-file toggle + env precedence, and a **secret-never-logged** assertion. An adversarial review (secret-leak / exec-safety / correctness) passed *fix-then-ship*; the one critical it found — a blocking `flock` on the exec path — is fixed in this branch.

## Notes / open decisions

- **CLI-only** for v1 (no GUI wiring); the JSON config file is deliberately placed where the app keeps its state, so a future desktop "Audit log: on/off" toggle can adopt it unchanged.
- Default is left **OFF** (logging key-names + argv + cwd of every pull by default is a privacy trade-off). Flipping to opt-out is a one-line change if preferred.
- Deferred to a follow-up: the optional daemon mirror (live-UI refresh + a daemon-`fsync`'d copy for gate/contain).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mxcl:
Kinda feel we should not make this optional

lofimichael:
> Kinda feel we should not make this optional

Agree - initial worry was about long-running ops bloating log sizes (and thus having it be opt-out), but I'm sure this is addressable if it comes up

lofimichael:
Agreed — made it always-on in 32fb800.

Dropped the opt-in/default-OFF config toggle, the `av audit enable`/`disable` commands, the persisted `audit-config.json`, and the `AV_AUDIT` on/off env var. The log is now written unconditionally (still best-effort, non-blocking, and exec-safe — `try_record` swallows every error so it can never delay or break a gated op).

Kept `AV_AUDIT_PATH` / `AV_AUDIT_NO_ARGV` / `AV_AUDIT_HMAC` / `AV_AUDIT_MAX_BYTES` — those tune *how* it records, not *whether*.

lofimichael:
fixed that @mxcl

mxcl:
Codex review:

---

**Findings**

- P1: HMAC setup breaks verification of existing chain-only records. Records written before `AV_AUDIT_HMAC=1` get `mac: ""` at [audit.rs](/private/tmp/automic-vault-pr37/src/lib/rs/audit.rs:331), but verification with a key requires every record to match a newly computed MAC at [audit.rs](/private/tmp/automic-vault-pr37/src/lib/rs/audit.rs:613). That contradicts the documented fallback at [audit.rs](/private/tmp/automic-vault-pr37/src/lib/rs/audit.rs:21). Repro path: use audit normally, later run `AV_AUDIT_HMAC=1 av audit setup`, then `AV_AUDIT_HMAC=1 av audit verify` fails on old records.

- P1: `AV_AUDIT_NO_ARGV` does not protect `av gate` messages. Redaction only clears `argv` and `cwd` at [audit.rs](/private/tmp/automic-vault-pr37/src/lib/rs/audit.rs:272), while `gate` writes the user-supplied message and denial reason verbatim at [gate.rs](/private/tmp/automic-vault-pr37/src/lib/rs/gate.rs:126). Since audit is now always-on, this can leak secrets typed into a gate message.

- P2: The pre-exec path is synchronous O(log size). Every record calls `read_tip`, which scans and JSON-parses the active audit log from the start at [audit.rs](/private/tmp/automic-vault-pr37/src/lib/rs/audit.rs:370). With an 8 MiB default rotation threshold, this can add latency before `exec`, so the “auditing never delays” contract is overstated. A tail read or locked sidecar tip would fit the design better.

**Open Question**

`AV_AUDIT_PATH` is honored directly at [config.rs](/private/tmp/automic-vault-pr37/src/lib/rs/config.rs:77), while audit write errors are swallowed. If “always on” means agents must not be able to bypass logging, this env var is effectively a bypass by pointing it at an unwritable or disposable path.

**Verification**

- `AV_COMBINED_DB_PATH=/private/tmp/automic-vault-pr37/src/lib/rs/fixtures/coverage-combined.json cargo test audit --lib`: passed, 12 tests.
- Full `cargo test --lib` fails in this worktree due package DB fixture/radioisotope expectations, not specifically the audit tests. Initial run also fails without `AV_COMBINED_DB_PATH` because `data/combined.json` is absent.
- `git diff --check origin/main...origin/pr-37`: clean.

I would not merge PR #37 until the HMAC mixed-history and gate-message redaction issues are fixed.

PR #56, opened 2026-07-18
Detect unsafe PATH precedence in shells and macOS GUI environments

mxcl:
## Summary

- Add shared detection for user-writable PATH entries ahead of protected system directories.
- Report unsafe PATH ordering in Bash, Zsh, and macOS launchd environments.
- Display the macOS GUI PATH in scan output and JSON reports.
- Document mitigation guidance and add detector, CLI, and path-security tests.

## Testing

- Added unit tests for PATH precedence, creatable and empty entries, macOS detection, and report output.
- Updated scan integration coverage for GUI PATH output and clean environments.
- Not run (not requested)


mxcl:
@copilot resolve your issues if you think they are now resolved pls

PR #72, opened 2026-07-24
Accept OSC-prefixed login-shell PATH output

### LINKED ISSUE #61: Doctor rejects a valid PATH when login-shell output has an OSC title prefix
Automic Vault 2.1.0 reports `Unable to inspect the login-shell PATH` even though the configured shell exits successfully, prints a valid PATH, and resolves `av` correctly.

## Setup

- macOS 26.5.2
- Automic Vault and `av` 2.1.0
- zsh 5.9
- Configured login shell: `/run/current-system/sw/bin/zsh`

## Reproduction

My interactive shell startup writes an OSC terminal-title sequence without a trailing newline. The same command used by `loginShellPATH()` therefore produces one line shaped like this:

```text
<ESC>]0;zsh<BEL>/path/one:/path/two:...
```

The shell itself succeeds:

```sh
/run/current-system/sw/bin/zsh -lic '/usr/bin/printenv PATH'
echo $?
# 0

command -v av
# /run/current-system/sw/bin/av

av doctor --json
# {"results":[]}
```

Refreshing Doctor in the app still reports:

```text
Unable to inspect the login-shell PATH
```

## What I found

`loginShellPATH()` runs the configured shell with `-lic /usr/bin/printenv PATH`, then accepts only a newline whose first character is `/`:

https://github.com/automic-vault/automic-vault/blob/3c85ddfc2085348c79f7f287346c6087f35b97db/src/menu-helper/Sources/MenubarHelperCore/DashboardSnapshot.swift#L1298-L1322

Because the OSC title bytes and PATH are on the same line, the valid PATH no longer satisfies `hasPrefix("/")`.

## Expected

Doctor recognizes the valid PATH from a successful login shell, or at least distinguishes unparseable decorated output from a shell startup failure.

## Observed

Doctor reports that it cannot inspect PATH and suggests fixing the login shell, even though the shell and CLI are both working. This does not appear to bypass a security boundary, but the false failure and `Install av CLI` action are misleading for an otherwise healthy setup.

mxcl:
Closes #61.

## What changed

- extract login-shell PATH parsing into a focused, testable function
- strip only leading OSC sequences terminated by BEL before applying the existing absolute-path check
- retain fail-closed behavior for unterminated OSC sequences and unrelated startup output

## Root cause

`loginShellPATH()` required the selected output line to start with `/`. A shell startup hook that emitted an OSC terminal-title sequence without a newline placed control bytes before an otherwise valid PATH, so Doctor treated the PATH as unavailable.

## Validation

```sh
swift test --package-path src/menu-helper --disable-automatic-resolution
```

40 tests passed.

### DIFF
--- a/src/menu-helper/Sources/MenubarHelperCore/DashboardSnapshot.swift
+++ b/src/menu-helper/Sources/MenubarHelperCore/DashboardSnapshot.swift
@@ -1322,7 +1322,15 @@ func loginShellPATH() -> String? {
 func loginShellPATH(from data: Data) -> String? {
     String(data: data, encoding: .utf8)?
         .split(whereSeparator: \.isNewline)
-        .last(where: { $0.hasPrefix("/") })
+        .compactMap { line -> Substring? in
+            var line = line
+            while line.hasPrefix("\u{1B}]") {
+                guard let terminator = line.firstIndex(of: "\u{07}") else { return nil }
+                line = line[line.index(after: terminator)...]
+            }
+            return line.hasPrefix("/") ? line : nil
+        }
+        .last
         .map(String.init)
 }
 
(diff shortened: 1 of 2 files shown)


mxcl:

PR #75, opened 2026-07-25
Fix in-app secret mutations to use direct authorization

### LINKED ISSUE #74: Can't add a secret in 2.2.0
After clicking "Save" to add a new secret in the UI, I get this dialog and the "Approve Once" button is not clickable. This seems like a regression in 2.2.0 only.

Image

A workaround is to use the av cli. But adding secret from the UI seems completely borked.

mxcl:
## Summary

- Route in-app secret mutations through the mutation’s direct `perform()` path.
- Avoid incorrectly invoking the approved-secret approval flow for in-app requests.
- Fixes #74

## Testing

- Tested both secret saving flows by human hands.

### DIFF
--- a/src/menu-helper/Sources/MenubarHelper/main.swift
+++ b/src/menu-helper/Sources/MenubarHelper/main.swift
@@ -1031,21 +1031,7 @@ private func performApprovedSecretMutation(
 func performInAppSecretMutation(
     _ mutation: SecretMutation
 ) -> (status: OSStatus?, error: String?) {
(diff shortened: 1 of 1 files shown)


mxcl:
It was not intentional to show the approval gate from inside the .app so this fixes that.

.apps have many protections that cli tools do not which influences our policy that the .app is a secure environment with the obvious caveat against tools being given Computer Use type approvals at the macOS TCC layer.
 
I would like to add a detector for Computer Use in future since it can be used to bypass our securities. But I also want to be able to move approval off the same computer (to phones, eg.) in future which would avoid this problem altogether.

PR #86, opened 2026-07-27
Report human approval decisions

mxcl:
## What changed

- Return the human's approval decision in the authenticated XPC reply.
- Print `automic vault: approved` or `automic vault: denied` after the prompt is resolved.
- Ignore unknown decision values and cover the mapping with a unit test.

## Why

`automic vault: human approval required` only indicates that the CLI is waiting. Without a final status, a later tool failure can be mistaken for an approval failure.

## Validation

- `cargo fmt --check`
- `cargo test --locked` (763 tests)
- `swift test --package-path src/menu-helper` (44 tests)


mxcl:
@copilot do you consider your issue resolved with the latest commit?

PR #97, opened 2026-07-30
Keep hardened Homebrew compatible with zsh compinit

### LINKED ISSUE #95: av harden homebrew chown to automic:vault breaks zsh's compinit security check
After running `sudo av harden homebrew`, zsh started printing this on every new shell:

```
zsh compinit: insecure directories and files, run compaudit for list.
Ignore insecure directories and files and continue [y] or abort compinit [n]?
```

`compaudit` confirms the affected paths are all under `/opt/homebrew/share/zsh`:

```
/opt/homebrew/share/zsh/site-functions
/opt/homebrew/share/zsh
/opt/homebrew/share/zsh/site-functions/_aws
/opt/homebrew/share/zsh/site-functions/_brew
... (full completion set for every installed formula)
```

`ls -ld` shows why: the hardener chowns these to `automic:vault`, matching what `homebrew.md` documents ("the root phase ... owns `/opt/homebrew` as `automic:vault`").

zsh's `compinit`/`compaudit` only trusts directories in `$fpath` owned by `root` or the invoking user, never a third-party account, regardless of permission bits (`755` here, not group/other-writable). So this is a false positive from a security standpoint, but zsh has no way to know that.

I checked the other hardeners (`aws_cli`, `gh_cli`, `stripe_cli`, `supabase`, `sudo`) and none of them chown anything under a directory that lives in a shell's `$fpath` — this is specific to `homebrew`'s tree-wide `chown` of `/opt/homebrew`, which happens to include `share/zsh/site-functions` (completions for every installed formula, not just Homebrew's own).

`src/isotopes/detectors/zsh/detector.md` documents a related but different issue (plaintext credentials in shell startup files) and explicitly invites opening an issue for "a safer integration," so filing this separately since it's a distinct interaction (compinit's ownership trust model vs. the homebrew hardener), not a credential-detection gap.

Notably, this is zsh-specific: bash's completion loader does no ownership audit at all, so the same `automic:vault` ownership causes no warning there.

Repro:
1. `sudo av harden homebrew` on Apple Silicon macOS with zsh as the default shell.
2. Open a new terminal / run `compaudit`.

If this is a fix you'd want, happy to send a PR for it, let me know if that works for you and which direction you'd prefer.

mxcl:
Fixes #95.

## What changed

- keep zsh completion directories, exposed links, and their in-prefix targets owned by the configured desktop account
- grant the `automic` account narrow inherited ACL access so Homebrew can continue updating completions
- repair newly created completion ownership after Brew exits using the existing explicit `sudo` transfer pattern
- reject completion links that resolve outside the Homebrew prefix
- warn during hardening and document that user-owned completion code is an intentional compatibility exception
- bump the Brew stub compatibility marker to V5 so existing installations receive the migration

## Why

The Homebrew hardener owns `/opt/homebrew` as `automic:vault`, but zsh `compaudit` trusts completion files only when root or the invoking desktop user owns them. That caused every shell startup to warn after hardening.

This keeps Homebrew running as the unprivileged `automic` service account rather than expanding the launcher to setuid-root. The security tradeoff is explicit: software running as the desktop account can modify zsh completion code, while the remainder of Homebrew remains protected.

## Validation

- `cargo test` (all tests pass)
- `cargo fmt --check`
- macOS ACL smoke test using the exact `automic` ACL
- zsh `compaudit` smoke test against a user-owned completion directory carrying that ACL

### DIFF
--- a/src/brew_stub/main.rs
+++ b/src/brew_stub/main.rs
@@ -1,15 +1,17 @@
 use std::ffi::{CStr, CString, OsString};
 use std::fs;
-use std::io::Read;
+use std::io::{self, Read};
 use std::os::unix::fs::{MetadataExt, OpenOptionsExt};
-use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-const MARKER: &str = "AUTOMIC_VAULT_BREW_STUB_V4";
+const MARKER: &str = "AUTOMIC_VAULT_BREW_STUB_V5";
 const TARGET: &str = "/opt/homebrew/bin/brew";
+const PREFIX: &str = "/opt/homebrew";
 const APPROVAL_SERVICE: &str = "com.automicvault.av2.approval";
 const CASK_USER_UID: &str = "/opt/homebrew/var/automic/cask-user-uid";
+const ZSH_COMPLETIONS: &str = "share/zsh/site-functions";
+const COMPLETION_ACL: &str = "user:automic allow read,write,execute,delete,append,readattr,writeattr,readextattr,writeextattr,readsecurity";
 
 #[derive(Debug, PartialEq, Eq)]
 struct AuthorizationRequest {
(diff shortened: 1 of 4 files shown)


mxcl:

PR #101, opened 2026-07-31
Support standalone Developer ID launchers

### LINKED ISSUE #96: Support explicit enrollment of standalone Developer ID-signed launchers
## Problem

Automic Vault currently recognizes persistent launcher identities only when an ancestor executable is contained inside a signed `.app` bundle.

Standalone agent CLIs such as Claude Code and Codex are valid Developer ID-signed, notarized Mach-O executables, but Automic discards their identities because their executable paths contain no `.app`.

This creates an inconsistent result:

- `av doctor claude` and `av doctor codex` report both standalone binaries as healthy and securely identifiable.
- The approval helper cannot use those same identities for persistent policies.
- When no bundle launcher is found, the fallback can identify Automic Vault itself because `av` is packaged inside `Automic Vault.app`.

I observed this with installed Automic Vault 2.2.1 and traced the materially unchanged attribution path on current source at `dee54dd4`.

## Where the identity is lost

In `src/menu-helper/Sources/MenubarHelper/main.swift:2892-2915`, Automic already obtains live signing information for every ancestor process:

- designated requirement
- Team ID
- signing identifier
- validity
- ad-hoc status

It then calls `appBundleURLs(containing:)`. If the executable is not inside a `.app`, that returns an empty list and the otherwise valid identity is discarded.

On this machine, Automic successfully extracts:

```text
Claude Code
identifier: com.anthropic.claude-code
team: Q6L2SF6YDW
ad-hoc: false
valid Developer ID designated requirement

Codex
identifier: codex
team: 2DC432GLL2
ad-hoc: false
valid Developer ID designated requirement
```

Security.framework supports this directly. `SecStaticCodeCreateWithPath`, `SecCodeCheckValidity`, and signing-information extraction all work for standalone Mach-O files without a bundle.

## Suggested design

Support explicit, human-driven enrollment of a validated standalone executable identity.

The existing policy schema already stores the security-bearing designated requirement, so this should not require a new authorization model.

Security constraints I would recommend:

1. Match the exact designated requirement, not Team ID alone.
2. Require a valid Developer ID signature with a signing identifier.
3. Reject ad-hoc and unsigned executables.
4. Reject Apple platform identities such as `/bin/zsh`.
5. Never automatically choose the nearest signed ancestor.
6. Keep enrollment as an explicit action in the Automic Vault app.
7. Revalidate the executable identity on every request.
8. Show the identifier, Team ID, resolved path, and complete requirement during enrollment.

Automatic selection would be unsafe. For many command-line invocations, the nearest signed ancestor can be `/bin/zsh`, which would collapse unrelated shell activity into one principal.

## Why this matters for launcher apps

This follows the discussion in https://github.com/automic-vault/automic-vault/issues/73.

I built https://github.com/kunchenguid/pi-launcher because Pi itself is ad-hoc signed. Claude Code and Codex are different: both already ship with valid vendor Developer ID signatures.

Wrapping and re-signing those binaries would replace the vendors' identities with my own signing identity, require ongoing re-notarization after upstream updates, and provide no additional cryptographic assurance.

I also measured the TCC responsible process for shell-executed Pi Launcher and standalone Claude on this setup. Both resolve to the same session supervisor. The separate TCC identity benefit of an app bundle appears to require LaunchServices launch semantics and does not apply to this terminal-preserving `exec` pattern.

Is there an internal security constraint that requires `.app` bundles despite the standalone executable already having a valid designated requirement? I can provide the read-only probes, complete ancestry walks, signing output, and TCC measurements, or test a branch.

mxcl:
## Summary

- recognize standalone Developer ID Application-signed executables in launcher ancestry
- require an exact enrolled designated requirement before standalone launchers receive an automatic policy
- allow signed executables in the launcher picker and show their identifier, Team ID, resolved path, and full requirement before enrollment
- apply standalone identities to secret gates, blessed scripts, and secret-name access without changing bundle default-policy behavior
- update the Claude and Codex doctors to verify their exact Developer ID identities and recommend their signed standalone/Homebrew installs
- document the verified Claude, Codex, and Pi distribution state, including Pi Launcher and self-signing options

## Security

Ad-hoc, unsigned, non-Developer-ID, and Apple platform executables remain ineligible. Every request still validates the live process signature, and persistence continues to match the exact designated requirement rather than Team ID or path.

The CLI doctors now require the expected signing identifier, Team ID, and Apple's Developer ID intermediate and leaf certificate markers. The current Claude and Codex release binaries were downloaded from their official distribution channels and passed those exact requirements. Pi v0.83.0 was checksum-verified from its official release and confirmed to be linker/ad-hoc signed with no Team ID; it remains rejected.

## Validation

- `cargo test --all-targets --all-features --locked` (771 unit tests plus integration targets)
- `swift test --package-path src/menu-helper` (50 tests)
- `AutomicVaultMenubar --self-check-standalone-launchers`
- direct `av doctor` checks against current official Claude and Codex macOS arm64 binaries
- `git diff --check`

Fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,14 +224,14 @@ Execution of blessed scripts *can* be configured to be automatically approved
 for specific launchers.
 
 > [!TIP]
-> Allowing specific launchers (ie. `.app`s) is powerful but requires careful
+> Allowing specific signed launchers is powerful but requires careful
 > consideration. Potential uses:
 >
 > - Keeping all hardened tools at read-only or lower and only defining use of
 >   those tools via immutable blessing is truly a level of safety developers
 >   have only ever dreamed of.
-> - Having a single terminal app that is the only used for deployments and the
->   the only `.app` that is endorsed to run them.
+> - Having a single terminal app that is the only one used for deployments and
+>   the only launcher that is endorsed to run them.
 > - Or conversely, trusting your agents more than your terminal where supply
 >   chain attacks are more likely to occur.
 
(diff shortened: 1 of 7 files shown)


mxcl:

PR #102, opened 2026-07-31
Require hardened launchers for new secret gate access

mxcl:
## Summary

- require Hardened Runtime before a launcher can be newly added to a secret gate
- allow the `allow-jit` and `allow-unsigned-executable-memory` compatibility exceptions used by JIT runtimes such as ChatGPT's bundled Codex launcher
- reject launchers that enable direct injection or debugging escapes through DYLD environment variables, disabled executable-page protection, disabled library validation, or debugger attachment
- persist the requirement on newly admitted policies and recheck the live launcher on every match
- leave existing stored launcher allowances grandfathered and unchanged

## Security model

Enrollment validates the app's strict nested code signature, Hardened Runtime flag, and entitlements before writing the allowance.

New policy records carry a `requiresHardenedRuntime` marker. At runtime, Automic reads the live ancestor executable's signature flags and entitlements. If its posture is no longer acceptable, the policy resolves to `no-access` even when the designated requirement still matches.

Older records have no marker and decode as grandfathered. Editing an existing allowance preserves that state; removing and re-adding it applies the new checks.

JIT executable-memory entitlements weaken W^X protections but do not by themselves permit another process to inject code or attach a debugger, so they remain compatible with admission. The following enabled entitlements prevent new secret-gate access:

- `com.apple.security.cs.allow-dyld-environment-variables`
- `com.apple.security.cs.disable-executable-page-protection`
- `com.apple.security.cs.disable-library-validation`
- `com.apple.security.get-task-allow`

## Validation

- `swift test --package-path src/menu-helper` (52 tests)
- `AutomicVaultMenubar --self-check-approvals`
- `cargo test --all-targets --all-features --locked` (771 unit tests plus integration targets)
- `git diff --check`

### DIFF
--- a/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
+++ b/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
@@ -555,7 +555,7 @@ final class DashboardModel: ObservableObject {
                 return
             }
             guard let signing = appBundleSigning(url) else {
-                self.errorMessage = "Could not read code signing identity for \(url.lastPathComponent)."
+                self.errorMessage = "Could not securely verify the code signature for \(url.lastPathComponent)."
                 return
             }
             guard signing.runtimeProtection.allowsSecretGateAccess else {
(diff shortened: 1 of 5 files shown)


mxcl:
@copilot do you consider your issues resolved

PR #105, opened 2026-07-31
Classify OpenSSH security-key handles as medium

### LINKED ISSUE #104: Possible false positive for openssh detector for security key based keypair
I have a lot of SSH keypairs on my machine. So Automic Vault reporting that...

> SSH private key is stored without passphrase encryption.

However, these keypairs are setup using `-sk` key type (`ecdsa-sk`, `ed25519-sk`), which basically storing the private key on the hardware security key and left the public key on the machine itself. 

To use the key, I have to enter the pin (which can be optional) then tap the hardware security key to authorize the keypair.
But, even though, I could configure the keypair setup to skip the tap and no pin required, the private key still stay on the hardware security key.

man [ssh-keygen](https://man7.org/linux/man-pages/man1/ssh-keygen.1.html#FIDO_AUTHENTICATOR)

- `verify-required` flag is needed for requiring pin
- `no-touch-required` flag can also be used to skip the tap

Perhaps we should lower the severity (since passphrase still somewhat useful) when the key type is security key backed. Alternatively, if we could approve certain keys to be safe/ignore by the detector, that fine as well.

mxcl:
Closes #104.

## What changed

- identify unencrypted `ecdsa-sk` and `ed25519-sk` key-handle files from matching public and private OpenSSH key records
- report those findings at medium severity with hardware-key-specific wording and remediation
- keep exportable, malformed, and inconsistent unencrypted private keys at high severity
- document the distinction and add focused coverage for both SK algorithms

## Why

The detector previously treated every unencrypted `openssh-key-v1` container as a complete exportable private key. FIDO SK files contain a key handle while the signing key remains on the authenticator, so they should not turn the menu-bar status red. A passphrase is still useful defense in depth, which is why the finding remains at medium severity.

## Validation

- `cargo test --all-targets --all-features --locked` (772 unit tests plus all integration targets)

### DIFF
--- a/src/isotopes/detectors/openssh/detector.md
+++ b/src/isotopes/detectors/openssh/detector.md
@@ -3,7 +3,7 @@
 ## Trigger Conditions
 
 - SSH private key is stored without passphrase encryption.
-- FIDO SSH key handle is stored without passphrase encryption (medium severity).
+- SSH security-key handle is stored without passphrase encryption (medium severity).
 
 ## Sensitive Files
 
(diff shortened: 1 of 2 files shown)


mxcl:
@copilot do you consider your issues resolved with the latest commit?

spywhere:
@mxcl yup, this looks good to me. Thanks for a swift fix.

PR #106, opened 2026-07-31
Add fail-closed Varlock dotenv blessings

mxcl:
Adds a signed `av dotenv resolve` path and persistent Dotenv blessings under the renamed Blessings UI.

Authorization binds the canonical root `.env.schema` SHA-256, direct static `av()` declaration, live parent process metadata, and signed launcher. Schema and ancestry are revalidated immediately before disclosure; failures deny without fallback.

Also adds Keychain persistence, stale blessing display/revocation, launcher configuration, and focused Swift/Rust checks.

Companion plugin: https://github.com/automic-vault/varlock-plugin

### DIFF
--- a/src/cli/dotenv.rs
+++ b/src/cli/dotenv.rs
@@ -1,0 +1,111 @@
+use std::ffi::OsString;
+use std::io::Write;
+use std::path::PathBuf;
+
+const USAGE: &str = "usage: av dotenv resolve --schema PATH --item NAME --key NAME";
+
+struct Options {
+    schema: PathBuf,
+    item: String,
+    key: String,
+}
+
+pub(super) fn run(args: Vec<OsString>, stdout: &mut dyn Write, stderr: &mut dyn Write) -> i32 {
+    let options = match parse(args) {
+        Ok(options) => options,
+        Err(err) => {
+            let _ = writeln!(stderr, "av dotenv: {err}\n{USAGE}");
+            return 2;
+        }
+    };
+    match crate::secrets::resolve_dotenv_secret(
+        &options.schema.to_string_lossy(),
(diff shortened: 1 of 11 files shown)


mxcl:
Pausing this PR while we redesign the integration.

Decisions for the next iteration:

- Remove `av dotenv resolve`. We do not want an endorsed CLI that emits plaintext secrets over stdout.
- Have the Varlock JavaScript package talk directly to Automic Vault through a minimal universal N-API bridge and a separate dotenv-only XPC service. The existing signed-client XPC service remains unchanged.
- The new service may receive requests from unsigned or differently signed hosts such as Node and Bun, so the approval UI must prominently identify the client as unverified and show the client path/signature status, verified launcher, process chain, schema path, and schema SHA-256.
- Treat the native bridge as transport, not authentication. Automic Vault must independently validate the peer PID, live launcher/entrypoint ancestry, exact root `.env.schema` path and hash, and the direct static `av()` declarations, then revalidate immediately before disclosure. Any ambiguity fails closed.
- Persistent approval is configured in the app like Blessed Scripts, phrased around blessing this dotenv for the verified launcher. It is not tied to an exact Automic Vault key set.
- Add a Keychain-backed “Varlock access” switch, default off. Disabling it stops access, clears one-time state, and invalidates existing dotenv blessings so re-enabling cannot silently restore trust.
- Plaintext will necessarily exist in the authorized language process memory. The security goal is to avoid providing a supported shell/stdio secret-dump primitive, not to claim authorized code cannot reveal its own secrets.

We will resume from this direction later; no further work is planned on this PR for now.

mxcl:
Closed in favor of #171

PR #119, opened 2026-08-04
Report every pending approval waiter

### LINKED ISSUE #54: Print approval-wait status for every blocked invocation
## Problem

When an approval prompt is already pending, subsequent commands waiting on human approval can appear to hang silently.

This was especially confusing from a Vaultty remote session to `pangolin`: repeated `aws s3 ls` invocations blocked as expected, but later invocations did not print:

```
automic vault: human approval required
```

The approval system itself was not stuck; the missing status made the remote commands look stuck.

## Reproduction

1. In a Vaultty remote session, run `aws s3 ls` so it reaches a human approval gate.
2. Before resolving the pending approval, invoke `aws s3 ls` again.
3. Observe that the later invocation waits without printing the human-approval-required message.

## Expected

Every invocation blocked waiting for a human decision should promptly print `automic vault: human approval required` to stderr, even when another approval request is already pending.

## Notes

A local invocation from Codex printed the status message correctly, so this may be specific to repeated or concurrent waiters, prompt serialization, or the Vaultty remote-session path rather than the basic XPC event handler.

mxcl:
## Summary

- send `human-approval-required` before entering the serialized AppKit approval queue
- make the transient approval cache safe for concurrent XPC handlers

## Root cause

The server emitted the status event from the main-queue block immediately before `NSApp.runModal`. While one prompt was active, later waiter blocks could not run, so their clients appeared to hang silently.

The status check now runs after caller validation and automatic-approval checks but before UI serialization. The final cache decision is still rechecked on the main path.

## Security

This does not change caller identity validation, gate policy, approval decisions, secret loading, or audit-before-release behavior. Cache access is serialized with `NSLock` to preserve its existing semantics across concurrent XPC handlers.

## Checks

- `cargo test`
- `swift test` (58 tests)
- `swift run AutomicVaultMenubar --self-check-transient-approvals`

Fixes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,15 @@ the approval level for each tool *based on the launcher*:
 Hardened tools request secrets when they need them. Automic Vault releases those
 secrets only when your gate policy allows it or you approve.
 
+> [!IMPORTANT]
+> Human approval is available only while your macOS user session is active and
+> the screens are awake. Automic Vault aborts open approval windows and denies
+> requests that still need a human decision when the session becomes inactive
+> or the screens sleep. Retry when you return; stale approvals are not preserved.
+>
+> Requests already allowed by policy may continue without a prompt, subject to
+> each secret’s **Available While Locked** setting.
+
 > [!NOTE]
 > For launcher-specific policy, we walk the process’s launcher chain, validate its code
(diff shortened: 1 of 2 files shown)


mxcl:

PR #126, opened 2026-08-05
Detect plaintext Codex CLI credentials

than:
Codex caches its login in a plaintext file at `${CODEX_HOME:-~/.codex}/auth.json` by default on every platform, macOS included. It's created `0600`, which stops other users but not anything running as the same user.

Worth flagging rather than accepting as unavoidable, because the contents are durable. Alongside an optional `OPENAI_API_KEY` and personal access token, the ChatGPT token set carries a **refresh token**, so a stolen copy keeps working after the access token expires.

This looked like the same class the tree already covers — `aws-cli-credentials-file` and `gh-cli-hosts-token` both flag `0600` files for containing a plaintext credential — and Codex looked like a gap, with `qwen-code` currently the only detector aimed at an AI coding tool.

### The remediation is Codex's own, so this is detect-only

`cli_auth_credentials_store = "keyring"` moves the credential into the system keyring, so `detector.md` documents that route instead of shipping a hardener. Three details are in the doc because they change the advice:

- The default really is `file` on macOS. Two upstream PRs that would have flipped it to keyring (openai/codex#11875, #11880) were closed unmerged, so this is current behavior rather than an old-version problem.
- Changing the setting does not migrate or delete the file already on disk, so `rm` is a required step (openai/codex#14704).
- `keyring` is the right recommendation over `auto` on a workstation: `auto` silently falls back to the plaintext file, `keyring` fails loudly.

### Verification

`cargo test --all-targets --all-features --locked` is green (781 lib tests). Nine unit tests cover API-key mode, ChatGPT token sets, a refresh token on its own, empty and whitespace values, non-object and unparseable JSON, a missing file, and the `CODEX_HOME` override.

Also ran the built binary against a real machine to confirm it fires and reports only the path:

```
├─ 7. codex
│  severity HIGH
│  problem
│  Codex CLI auth file contains plaintext credentials.
│  affected files
│  • /Users/…/.codex/auth.json
```

### Two things I did not want to guess at

1. `radioisotope::docs_url` has no `"codex"` arm, so `homepage` falls back to the `radioisotopes` repo root while every other detector points at `radioisotopes/<name>`. I left the fallback rather than link a directory that doesn't exist yet — happy to add the arm if you'd like a `radioisotopes/codex` entry to go with it.
2. `cargo fmt --check` reports a diff at `tests/release_workflow.rs:162` on an untouched `main` as well as on this branch, so it looks pre-existing rather than something this PR introduces. Left alone. It may just be that my rustfmt (1.9.0) is newer than the one CI uses.

### DIFF
--- a/src/isotopes/detectors/codex/detector.md
+++ b/src/isotopes/detectors/codex/detector.md
@@ -1,0 +1,40 @@
+# codex Detector
+
+## Trigger Conditions
+
+- Codex CLI auth file contains a plaintext API key or ChatGPT token set.
+
+## Sensitive Files
+
+- `${CODEX_HOME:-$HOME/.codex}/auth.json`
+
+## Why This Matters
+
+Codex caches login credentials in a plaintext file by default on every platform,
+including macOS. The file is created mode `0600`, which stops other users but not
+anything running as you. It holds a refresh token alongside any API key, so a
+copy keeps working after the access token expires.
+
+## Why This is not Yet Hardened
+
+Codex can store credentials in the system keyring itself, so the fix belongs in
+its configuration rather than in a wrapper. Set `cli_auth_credentials_store` to
+`keyring` in `${CODEX_HOME:-$HOME/.codex}/config.toml`, run `codex login` again,
+then delete the plaintext file left behind.
+
+```toml
+cli_auth_credentials_store = "keyring"
+```
+
+```sh
+codex login
+rm ~/.codex/auth.json
+```
+
+The last step matters. Changing the setting neither migrates nor removes the file
+already on disk, so the plaintext copy survives until you delete it.
+
+Prefer `keyring` over `auto` on a workstation. `auto` falls back to the plaintext
+file when no keyring is available, while `keyring` fails loudly.
+
+[Open an issue to discuss a safer integration](https://github.com/automic-vault/automic-vault/issues).
--- a/src/isotopes/detectors/codex/detector.md
+++ b/src/isotopes/detectors/codex/detector.md
@@ -2,7 +2,9 @@
 
 ## Trigger Conditions
 
-- Codex CLI auth file contains a plaintext API key or ChatGPT token set.
+- Codex CLI auth file contains a plaintext API key, personal access token,
+  ChatGPT token set, Bedrock API key, or agent identity.
+- Codex CLI auth file exists but cannot be read or parsed.
 
 ## Sensitive Files
 
@@ -28,7 +30,7 @@ cli_auth_credentials_store = "keyring"
 
 ```sh
 codex login
-rm ~/.codex/auth.json
+rm "${CODEX_HOME:-$HOME/.codex}/auth.json"
 ```
 
 The last step matters. Changing the setting neither migrates nor removes the file
--- a/src/isotopes/detectors/codex/mod.rs
+++ b/src/isotopes/detectors/codex/mod.rs
@@ -1,0 +1,186 @@
+#![allow(dead_code)]
+
+use std::path::{Path, PathBuf};
+
+pub fn install_is_insecure() -> Result<bool, String> {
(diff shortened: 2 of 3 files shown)


mxcl:
Thanks for adding this. I ran `cargo test --all-targets --all-features --locked`; all tests passed.

I found three security issues that I think need fixing before merge:

1. **The detector misses current credential shapes.** `bedrock_api_key` is stored as an object, so passing it through `is_nonempty_string` always returns false. Current Codex auth can also contain `agent_identity` credentials, which this detector does not inspect. Please handle both structured forms and add tests using Codex's current schema: https://github.com/openai/codex/blob/2707dfc219b9bc9d38d7d37d1f691855f7b44c1e/codex-rs/login/src/auth/storage.rs#L38-L76

2. **Malformed JSON fails open in the real scan path.** `auth_file_has_credentials` returns an error, but `radioisotope::findings` converts detector errors into an empty finding list. The test named `reports_unparseable_json_rather_than_staying_silent` therefore does not cover product behavior. Please make the scan surface a finding for an unreadable credential file and test through `findings`, not only the parser helper.

3. **The remediation command ignores `CODEX_HOME`.** The detector and prose resolve `${CODEX_HOME:-$HOME/.codex}`, but the command deletes `~/.codex/auth.json`. A user with a custom Codex home would leave the reported plaintext credential in place. Please delete the same resolved path and quote it safely.

Two non-blocking cleanup requests:

- Mention `personal_access_token` in the trigger documentation, since the code detects it.
- Avoid uncoordinated global `HOME`/`CODEX_HOME` mutation in the tests; those tests can race when the test runner executes them concurrently.

mxcl:
On the two things you held back:

1. **Leave `radioisotope::docs_url` alone in this PR.** The radioisotopes repository no longer exists, so adding a `"codex"` match arm would create another dead URL. The registered detector pass overwrites this legacy URL with this repository's `src/isotopes/detectors/codex/detector.md` URL before returning findings, so users do not receive the dead fallback. Cleaning up the legacy mapping is separate repo-wide work, not something this PR needs to solve.

2. **Your rustfmt diagnosis is correct.** Current `main` fails `cargo fmt --all -- --check` on the same untouched line, and PR CI inherits that failure. Please do not fold the unrelated formatting change into this branch. We'll fix it separately in #128: https://github.com/automic-vault/automic-vault/issues/128

No changes requested from you for either item.

than:
Thanks, all three were real. Fixed and pushed (also rebased onto current `main`).

**1. Structured credential shapes.** You're right that `bedrock_api_key` never matched — it's an object, so `is_nonempty_string` was always false. Same for `agent_identity`, which is an untagged enum: either a bare JWT string or a record carrying `agent_private_key`. Rather than enumerate the current layout, detection now walks whatever structure is present and looks for non-empty string material inside it, which should survive the next schema change too. Tests cover the object form of `bedrock_api_key`, and both the JWT and record forms of `agent_identity`, using the schema you linked.

**2. Failing open on malformed JSON.** This was the worst of the three, and you're right that my test proved nothing about product behavior. `insecurity_reasons_for` now returns a finding when the file can't be read or parsed, instead of an `Err` the scan throws away — an unreadable credential file isn't evidence of safety. There are tests through `findings` for both the unparseable and the missing-file cases.

**3. `CODEX_HOME` in the remediation.** Fixed, now `rm "${CODEX_HOME:-$HOME/.codex}/auth.json"`.

Both cleanups are in as well: trigger conditions mention `personal_access_token` (and the other shapes), and path resolution moved into a pure function taking the two env values as arguments, so no test mutates `HOME` or `CODEX_HOME` any more.

Thanks also for the note on `docs_url` — I didn't know the radioisotopes repo was gone, and it's good to know the registered pass overwrites the legacy URL anyway. Left it alone, and left the formatting to #128.

`cargo test --all-targets --all-features --locked` is green at 791, and `cargo fmt --all -- --check` is clean on this branch now that #128 has landed on main.

PR #132, opened 2026-08-05
Open settings with Command-comma

than:
⌘, opens settings in essentially every other Mac app, and right now it does nothing here. The window already answers ⌘C, ⌘W and ⌘H in `performKeyEquivalent`, so this adds a fourth case alongside them. Since the app runs as an accessory with no menu bar to carry the standard Settings item, the window seemed like the right place for it rather than a menu.

Three small additions, each mirroring the existing `showSecretGate(id:)` path:

- `case ","` in `AutomicVaultWindow.performKeyEquivalent`
- `AutomicVaultMainWindowController.showSettings()`
- `DashboardModel.showSettings()`, setting `selectedSection = .settings` and clearing `selectedItemID`

### One thing I couldn't verify

It builds clean — `swift build -c release` on the whole package with Xcode 27 beta — and `swift test` is green at 60. But I couldn't confirm the pane actually switches at runtime, because I can't produce a signed build: `ApprovalServer.init` requires a team identifier, and running the bare SwiftPM executable exits immediately since it has no bundle identifier. Would you mind giving it a quick try?

If it doesn't behave, my guess at the likely culprit is focus: if the sidebar search field has first responder, the field editor may consume the event before the window sees it. Happy to move the handling if so.

### A question on scope

Right now this only works when the main window is open and key. Would you want ⌘, to also open the window when it's closed — say from the menu bar item — or is window-scoped the right behavior for an accessory app?

### DIFF
--- a/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
+++ b/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
@@ -66,6 +66,10 @@ final class AutomicVaultMainWindowController: NSHostingController<DashboardRootV
         model.showSecretGate(id: id)
     }
 
+    func showSettings() {
+        model.showSettings()
+    }
+
     func reviewBlessing(
         _ request: BlessedScriptReviewRequest,
         completion: @escaping (String?) -> Void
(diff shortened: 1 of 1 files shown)


mxcl:
Yeah this to some extent doesn't work because we are just a menu bar app. When you have a “full” app you get these for free.

I intend to make us a full app at some point, it adds complexity because then we have to communicate state between the helper (the menu bar app) and the main app hence not done it yet. I want to be solid AF first.

This is a welcome fix to bridge the gap in the mean time! Thank you.

Issue #73, opened 2026-07-25
Persistent policy for a signed .app without a hardened-tool gate


kunchenguid:
Running Automic Vault 2.1.0.

I signed my own launcher app so my Pi coding agents run under a stable identity, and the attribution side works exactly as I hoped. Running `av inject +HCLOUD_TOKEN -- sh -c '...'` from inside it gives me a prompt that correctly names my app (`com.kunchenguid.pi-launcher`, Developer ID team `9T2J7MNUP9`), not `sh`, not `av`. That's great.

The problem is I can only ever click Approve Once. Always Approve needs an existing Secret Gate, and gates only come from your compiled hardeners. I don't have any hardened tools, so there's no gate to attach my app to, and no way to create one. So a correctly recognized, notarized, Developer ID app can never get a persistent policy for a direct `av inject`.

I don't want to harden an unrelated tool just to manufacture a gate. The node hardener also doesn't quite fit my setup (nix + nvm + corepack).

What I'd like: let me grant a scoped, persistent policy to a signed `.app` for `av inject` directly for certain secrets - app identity plus a key pattern would be enough.

mxcl:
Yep agreed. I started off with a tight scope to get the features right and it's time to expand it more broadly. It requires some thought to do in a way that doesn't invalidate our product security promises. So I'll give it some thought and then reply here.

mxcl:
The revised blessing system may work for you for now (in v2.3.0). It allows blessing a script and auto-approving it for specific launchers.

I will make it so blessings can also apply to bundles (like .apps). Secret gates have granular access controls, being the main distinction (as well as internal reasons).

I agree the UI is a little muddled and I think we can combine these 2 things at some point in the future when we have seen wider usage and got some more user-stories from real users.

Allowing bundles to be blessed is a very good idea IMO since bundles get separate macOS TCC gates that mean you are separating privileges and not giving Pi the same access controls as your terminal. So this needs to be a dedicated support route for sure. Blessed bundles won’t require as many restrictions as scripts (I will audit this assumption) since the code signing system gives us various guarantees.

## Caveats

Signing a bundle properly requires a $100 / yr sub so I hope we can also support ad-hoc signed bundles, with thus the same restrictions as blessed scripts. I am not sure if this works well though. It's quite possible adhoc signed bundles bypass some or all TCC gates.

# Questions I Have

- How did you make the launcher app may I ask?

mxcl:
[...]

PR #136, opened 2026-08-07
Support hardening without Homebrew

mxcl:
## What changed

- Resolve simple hardener targets from PATH and embed the confirmed executable target in the installed stub.
- Reject an existing launcher stub that points its embedded target back to itself.
- Use the Automic Vault Homebrew tap automatically when Homebrew is available.
- Install supported isotope binaries in /usr/local/bin when Homebrew is unavailable.
- Download isotope metadata and archives in-process with ureq and Rustls, HTTPS-only redirects, and bounded timeouts.
- Report direct-install isotope updates through doctor and validate PATH candidates with code-signing checks.
- Remove the top-level Homebrew requirement from the README.

## Why

Nix and other non-Homebrew users should be able to harden existing tools without adopting a second package manager. Homebrew users should not need to type installation pre-steps manually.

## Impact

The hardening flow now previews the concrete installation or stub target, asks for confirmation, and performs the selected installation. Homebrew-managed isotopes continue to rely on Homebrew for updates; direct installs are updated by running av harden again.

## Validation

- cargo test --workspace
- 822 library tests plus CLI and integration tests passed
- Focused HTTPS-only and recursive-stub regression tests passed
- cargo fmt --check
- git diff --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,69 @@ version = "3.0.0"
 dependencies = [
  "libc",
  "serde_json",
+ "ureq",
 ]
+
+[[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
+name = "bytes"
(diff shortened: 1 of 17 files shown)


mxcl:
Refs #73

mxcl:

Issue #137, opened 2026-08-07
Bundle unsigned CLI tools as Secure Launcher apps


mxcl:
## Problem

Unsigned and arbitrary ad-hoc-signed CLI tools cannot safely serve as persistent Launcher identities. Their paths are mutable, they cannot provide a stable publisher identity, and without Hardened Runtime other same-user processes may be able to inspect or alter them.

This surfaced in #73 for tools such as `herdr`: when a detached process becomes the attributed Launcher, it cannot qualify as a Verified Launcher and requests fall back to Approval.

## Proposal

Let Automic Vault create a **Secure Launcher** app around one selected CLI payload. The generated app should:

- contain a minimal launcher that remains the payload's parent process;
- execute one fixed bundled payload rather than resolve a command through `PATH`;
- seal the payload and bundle resources;
- sign nested code from the inside out;
- enable Hardened Runtime with only the entitlements the payload demonstrably needs; and
- bind enrollment to the generated bundle, its designated requirement, and the exact payload digest.

Automic-created ad-hoc signing can be the default. Automic Vault must recognize only bundles it generated and explicitly enrolled—not arbitrary ad-hoc apps chosen from disk. If the user selects a Developer ID Application identity, the payload must remain pinned because the user, not the CLI's publisher, supplied that identity.

Once enrolled and verified, the Secure Launcher can participate in Authorization Gate policy like another Verified Launcher. Launching the original unbundled executable remains outside that policy.

## Security requirements

- Updates fail closed: any payload or sealed-resource change invalidates the enrolled identity.
- Re-enrollment after an update is attended and explicit.
- The CLI or agent cannot authorize, bless, or enroll its own replacement.
- A stable Developer ID signature must not silently carry policy to changed payload contents.
- Missing identity, integrity, payload, entitlement, or enrollment evidence denies access; failures must not broaden authority.
- Product copy must explain that sealing and Hardened Runtime reduce mutation and injection risk but do not make mutable prompts, configuration, plugins, extensions, or the payload itself trustworthy.
- New Secure Launchers start with conservative Access Levels.

## Acceptance criteria

- A user can select an unsigned or ad-hoc-signed standalone CLI and create a Secure Launcher app.
- The app launches the pinned payload with arguments while preserving Launcher attribution.
- Automic Vault can enroll and reverify the generated app as a Verified Launcher.
- The generated app can be assigned policy at an applicable Authorization Gate.
- Tampering with or replacing the payload causes verification to fail closed.
[...]

PR #138, opened 2026-08-07
Add biometric Full Access Sessions

robinbraemer:
## Summary

- add a Touch ID-only **Full Access Session** that automically authorizes every valid recognized operation for eligible verified launchers across Authorization Gates
- let the user choose **1 hour**, **8 hours**, or **Until turned off**, with Until turned off selected by default
- keep the session and its selected lifetime in memory only; timed sessions use wall-clock and monotonic deadlines, and every session ends on lock, display sleep, app exit/update, or explicit user action
- add conspicuous menu-bar, banner, remaining-time/Until-turned-off, Settings, and Authorization History presentation
- preserve fail-closed behavior for Unknown operations, unverifiable launchers, untrusted Gate Clients or Targets, malformed requests, missing Secrets, and failed required Authorization Records
- document the security boundary and decision in the canonical domain language, architecture, and ADR 0006

## Security properties

- activation calls LocalAuthentication with `deviceOwnerAuthenticationWithBiometrics`; password fallback is not accepted and biometric reuse duration is zero
- the selected lifetime is part of the local Touch ID activation flow; no CLI, URL, XPC, environment, preference, or persisted activation interface exists
- pending biometric attempts are invalidated on End, lock, or display sleep and rechecked after authentication before the session starts
- automatic release holds a generation-bound session lease across payload loading, required audit persistence, provenance recording, and XPC reply, so End/lock/sleep cannot race with release
- explicit per-launcher Approval Required policies are temporarily overlaid only when the launcher remains eligible; failed hardened-runtime requirements are not bypassed
- an active Full Access Session outranks a narrower active script Blessing; normal durable policy retains the existing Blessing boundary
- Unknown classifications never receive Full Access
- the existing record-before-release path remains mandatory, with `Full Access Session` recorded as the policy reason
- choosing Install and Relaunch revokes active and pending sessions before update UI hides the End control
- **Until turned off** is not persisted and does not survive lock, display sleep, app exit/update, or an explicit End action

## Verification

- `swift build --package-path src/menu-helper --disable-automatic-resolution --product AutomicVaultMenubar`
- `swift build --package-path src/menu-helper --disable-automatic-resolution -c release --product AutomicVaultMenubar`
- `swift test --disable-automatic-resolution` with local Swift Testing framework search/rpath flags (81 tests)
- `AutomicVaultMenubar --self-check-full-access-session`
- `AutomicVaultMenubar --self-check-dashboard-search`
- `AutomicVaultMenubar --self-check-menu-status`
- `cargo test --all-targets --all-features --locked` (817 library tests plus binary and integration suites)
- `git diff --check`

### DIFF
--- a/docs/adr/0006-biometric-full-access-sessions.md
+++ b/docs/adr/0006-biometric-full-access-sessions.md
@@ -15,21 +15,21 @@ Add an in-memory **Full Access Session**. While active, the policy resolver appl
 
 Only the signed macOS app UI can start a session. Starting requires `deviceOwnerAuthenticationWithBiometrics` through LocalAuthentication, with no password fallback. Authentication cancellation, failure, lockout, or unavailable biometry leaves normal policy active. Gate Clients, Launchers, CLI commands, URL handlers, XPC messages, environment variables, and persisted configuration receive no start mechanism.
 
-A session lasts for at most one hour. It ends earlier when the macOS user session becomes inactive, the displays sleep, the app exits or updates, or the user selects **End Full Access Session**. Ending a session is always available without authentication because it narrows authority. The session is never persisted; app launch always starts under durable policy.
+The user chooses a session lifetime of **1 hour**, **8 hours**, or **Until turned off** before authenticating. Until turned off is the default. Timed sessions expire using both wall-clock and monotonic deadlines. Every session ends when the macOS user session becomes inactive, the displays sleep, the app exits or updates, or the user selects **End Full Access Session**. Ending a session is always available without authentication because it narrows authority. The session and selected lifetime are never persisted; app launch always starts under durable policy with Until turned off selected.
 
 Make the exceptional state conspicuous:
 
-- the menu-bar icon and menu use a warning treatment and show the remaining time;
+- the menu-bar icon and menu use a warning treatment and show the remaining time or Until turned off;
 - the main window shows a persistent warning banner with an End action;
 - Settings explains exactly what Full Access includes and what still fails closed;
 - Authorization History records allowed requests with `Full Access Session` as the policy reason.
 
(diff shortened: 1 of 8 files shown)


mxcl:
I like the thinking here. Let me review.

mxcl:
K I've experimented with this and I feel it is a little overly broad. I would prefer being able to upgrade specific launchers to have full access for a smaller time window (or perhaps returning a token to the launcher that if set as an ENV variable allows full access for a time window).

May I ask what prompted your desire to elevate every Tool’s capaibilities to full for *all* launchers? If I can understand the use case I can better design the product.

NOTE: It’s not that I'm saying we cannot have this feature, I just want to understand your user story first to make sure we do this right.

mxcl:
Well my question is more what examples of every 10 seconds are you getting? The way I have it configured I get approval prompts maybe once an hour.

Have you got your authorization gates set up to reasonable levels? What agents are you using etc.?

Like it's possible you need #137 because you are using agents without code signing and thus you cannot set up your Authorization Gates to be reasonable for the agent.

Or you need a few blessed scripts perhaps.

Having said this I think a “authorize aws for full access for the next 5 minutes” is a very reasonable thing to add, but just trying to see what sort of things are making you think that is not enough before we design these features.

The goal with Automic Vault is to give you security with as little friction as possible. But security will always have a cost and it's possible we can't meet you there. But I don't think so. I think we can. I just need to understand better what is going on for you so I can either: improve the docs or improve the tool.

robinbraemer:
Most prominently its `gh` commands: pr, checks runs, watch, merge, ...

mxcl:
Superseded by #168, which implements a fixed 10-minute Write Access grant scoped to one Tool-specific Authorization Gate, one Verified Launcher/runtime posture, and one Codex or Claude task context, with persistent on-screen visibility and an immediate End action.

mxcl:
If this was triggered by getting approval prompts for eg. every time you execute `npm` that will be fixed in the next release. These hardened tools were unnecessarily injecting secrets for too many subcommands.

Issue #141, opened 2026-08-09
Docs: clarify App Access override matching (bundle ID vs. Team ID) — especially for bare CLI executables outside /Applications


nbost130:
## What's unclear

A Secret Gate's "App Access" list lets you add a specific app as an override (e.g. we added Warp — bundle id `dev.warp.Warp-Stable`, Team `2BBY89MBSN` — at Full Access, and Warp-launched calls auto-approve cleanly as expected).

We wanted to add **Claude Code** (the CLI coding agent) as a trusted launcher, since its calls weren't auto-approving even with the gate's default "All Other Apps" row already set to Full Access. `codesign -dv` confirmed the actual binary is:

- Path: `/Users/nbost/.local/share/claude/versions/2.1.226` — a bare executable, **not** inside an `.app` bundle, **not** in `/Applications`
- `Identifier=com.anthropic.claude-code`
- `TeamIdentifier=Q6L2SF6YDW`

The **+** "Add App Access" picker only browses `.app` bundles under `/Applications` — it can't navigate to or select a bare executable living elsewhere. So we couldn't add the CLI binary directly.

As a workaround we added the separate **"Claude" desktop app** instead (a different Anthropic product, different bundle identifier — something like `com.anthropic.claudefordesktop` — but the *same* Developer Team ID, `Q6L2SF6YDW`). That entry showed `Signing identity u...` (truncated in the UI — unclear what it said) rather than a clean `Team <ID>` line the way Warp's entry displays.

After adding it, Claude Code CLI calls that were previously prompting started auto-approving. **We don't know why** — possibilities:
1. App Access matching is scoped by Team ID alone, so any app signed by `Q6L2SF6YDW` (desktop app or CLI) is now trusted under this override, or
2. Something else changed coincidentally (approval history accumulating, a different invocation shape resolving differently, etc.)

## The ask

Could the docs (or the app's own UI copy) clarify:

1. **Is App Access override matching scoped by `(bundle_id, team_id)` together, or by `team_id` alone?** This materially changes what adding an app actually grants — a broad team-level grant vs. a narrow single-bundle grant is a very different security posture, and right now there's no way to tell which one you're getting.
[...]

PR #143, opened 2026-08-09
Allow version-numbered launcher executables

mxcl:
## Summary

- allow the launcher picker to select version-numbered executables such as Claude Code's `2.1.226`
- keep executable, strict code-signature, Developer ID, and Hardened Runtime validation as the admission boundary
- document path navigation, exact designated-requirement matching, and the separate purpose of `av bless --endorse-launcher`

## Root cause

macOS infers a dynamic data type from a version-number filename extension, so `NSOpenPanel` filtered the executable out before Automic Vault could validate its signature.

## Validation

- `swift test --package-path src/menu-helper --disable-automatic-resolution`
- `AutomicVaultMenubar --self-check-standalone-launchers`
- `cargo test --all-targets --all-features --locked`

Refs #141

### DIFF
--- a/docs/signed-cli-launchers.md
+++ b/docs/signed-cli-launchers.md
@@ -33,8 +33,8 @@ a version-number filename extension as data. Selection alone grants nothing:
 Automic Vault resolves symlinks and verifies that the selected file is executable
 and has an eligible signature. If the signature is missing, ad-hoc, invalid, or
 not Developer ID for a standalone executable, selection fails. If the identity
-cannot be verified later, automic authorization fails closed and requires
-Approval.
+cannot be verified later, Automic authorization fails closed and requires
+manual approval.
 
 Launcher-specific rules match the exact designated requirement, which binds the
 signing identifier and Team ID together. A rule does not match a sibling product
--- a/docs/signed-cli-launchers.md
+++ b/docs/signed-cli-launchers.md
@@ -23,13 +23,25 @@ Placing such an executable inside an unsigned app does not make it eligible.
 2. In Automic Vault Settings, add a launcher to the relevant tool or blessed
    script policy.
 3. Select the resolved native executable, not a shell or package-manager shim.
+   The picker starts in `/Applications`; press Command-Shift-G to enter another
+   path directly. Version-numbered executables such as `2.1.226` are supported.
 4. Review the identifier, Team ID, path, and designated requirement before
    allowing it.
 
-Automic Vault resolves symlinks and verifies the selected executable itself. If
-the signature is missing, ad-hoc, invalid, or not Developer ID for a standalone
-executable, selection fails. If the identity cannot be verified later, automatic
-approval fails closed and requires manual approval.
+The picker permits generic files because macOS may classify an executable with
+a version-number filename extension as data. Selection alone grants nothing:
+Automic Vault resolves symlinks and verifies that the selected file is executable
+and has an eligible signature. If the signature is missing, ad-hoc, invalid, or
+not Developer ID for a standalone executable, selection fails. If the identity
+cannot be verified later, automic authorization fails closed and requires
+Approval.
+
+Launcher-specific rules match the exact designated requirement, which binds the
+signing identifier and Team ID together. A rule does not match a sibling product
+from the same signing team, and it is not bound to the selected path. The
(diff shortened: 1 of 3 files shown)


nbost130:
Heads up — I think this issue and #102 collide for Claude Code specifically, and thought it was worth flagging since neither references the other.

## The collision

#102 ("Require hardened launchers for new secret gate access") rejects any launcher carrying `com.apple.security.cs.disable-library-validation`, with `allow-jit`/`allow-unsigned-executable-memory` explicitly carved out for "JIT runtimes such as ChatGPT's bundled Codex launcher."

Claude Code's actual entitlements (confirmed via `codesign -d --entitlements -` on the current official installer binary):

```
com.apple.security.cs.allow-jit                        = true   (allowed by #102's carve-out)
com.apple.security.cs.allow-unsigned-executable-memory  = true   (allowed by #102's carve-out)
com.apple.security.cs.disable-library-validation        = true   (NOT allowed — #102's blocking list)
```

So Claude Code has *both* the entitlements #102 already accommodates *and* the one it still blocks. Right now, trying to add it as a Secret Gate App Access override gives a hard rejection: *"Launcher cannot be allowed — com.anthropic.claude-code weakens Hardened Runtime with com.apple.security.cs.disable-library-validation and cannot receive secret-gate access."*

This issue (#143) fixes the picker so a version-numbered binary like Claude Code's `2.1.226` becomes *selectable* — but as far as I can tell, selecting it will still hit #102's rejection immediately after. If that's right, #143 alone won't actually let anyone enroll Claude Code as a launcher; it'll just move the failure one step later in the flow.

(Also — separate minor thing, in case it's not already known: `disable-library-validation` is almost certainly there to support MCP servers / native module loading, which is core to how Claude Code's plugin architecture works, not an oversight. So this isn't likely to go away on Claude Code's side.)

## Something that might help resolve it

Today I found that a **blessed script's own "Calling Apps" list does NOT enforce the #102 entitlement check** — I was able to endorse Claude Code as a Calling App on a blessed script (`av bless`'d, then added via the script's own detail view) and it shows a clean "Verified designated requirement," no rejection, even though the exact same binary gets hard-rejected when added to a Secret Gate's App Access list directly.

So there's already a precedent inside the app for treating "trusted to invoke this one specific reviewed script" as safe even for a launcher #102 considers too risky for broad gate access. Might be a useful model for #102/#143 to reconcile around — e.g. scoped trust (per-script) staying strict-optional the way it already is, while broad launcher trust (per-gate) keeps #102's current bar.

Happy to provide more detail (full entitlements dump, exact repro steps) if useful.

nbost130:
Quick follow-up for anyone landing here from the #102 collision noted above: found a working path *today* to run Claude Code through hardened tools without it ever needing Secret-Gate-level launcher trust — details in #73. Short version: a blessed script's own Calling-Apps list isn't subject to #102's entitlement check, and adding a `capabilities:` manifest comment after the shebang is what makes it fully silent instead of still prompting once per command shape. Doesn't resolve the #102/#143 collision itself, but it means nobody needs to wait on that resolution to get Claude Code working unattended.

mxcl:
Thanks for catching this. You're right that #143 alone would only move the Claude Code failure from the picker to the runtime-protection check.

We decided Claude Code should work as a first-class Verified Launcher. I folded the policy change into #142 and closed the separate #144, while keeping this PR focused on the version-number picker bug.

#142 now treats disabled library validation as a supported Hardened Runtime compatibility exception and warns that third-party libraries or plug-ins can inherit the Launcher's authority. New launcher rules store the accepted runtime posture and recheck it on every request, so an existing strict rule does not silently expand if its Launcher later disables library validation. Missing Hardened Runtime, DYLD environment-variable injection, disabled executable-page protection, and debugger attachment remain blocked.

With both PRs, #143 makes Claude Code's version-numbered executable selectable, and #142 lets the eligible signed executable receive an exact rule or the gate default. Identity still uses exact designated-requirement equality; there is no Anthropic-specific, path-based, or Team-ID-only exception.

Blessed Scripts remain the narrower option when that fits the workflow, but they will no longer be the only way to use Claude Code.

nbost130:
Thanks for being so responsive. Enjoying Automic so far and great to see
how active the project is.

PR #144, opened 2026-08-09
Support Claude Code's Hardened Runtime profile

mxcl:
## Summary

- support the Hardened Runtime profile used by Claude Code without adding an Anthropic-specific identity exception
- treat disabled library validation as an eligible, warned compatibility exception while continuing to block injection, executable-page, debugger, and missing-runtime failures
- persist the accepted runtime profile on new Secret Gate and Direct Access rules and revalidate the live Launcher on every request
- preserve strict existing rules, grandfathered pre-runtime policies, exact designated-requirement matching, and fail-closed behavior

## Why

PR #143 makes Claude Code version-numbered executables selectable, but #102 still rejects the official binary because it carries `com.apple.security.cs.disable-library-validation`. Selection alone therefore does not resolve the Claude Code use case in #141.

Apple distinguishes disabled library validation, which permits a host to load third-party code, from DYLD environment-variable injection. This PR preserves that distinction while warning that loaded libraries and plug-ins inherit the Launcher authority.

## Security model

New durable rules record one of two checked postures:

- strict Hardened Runtime
- Hardened Runtime allowing disabled library validation

A rule accepts the recorded posture or a stronger one. Removing the library-validation exception remains valid; adding it to a strict Launcher invalidates automic authorization until the user creates a new rule. Missing Hardened Runtime and these exceptions remain blocked:

- `com.apple.security.cs.allow-dyld-environment-variables`
- `com.apple.security.cs.disable-executable-page-protection`
- `com.apple.security.get-task-allow`

Existing #102 records decode as strict. Older deliberately grandfathered policies remain grandfathered. Existing Direct Access Rules decode as strict. The additive stored field is rollback-safe: older builds ignore it, retain the compatibility Boolean, and continue to reject disabled library validation.

The picker and Direct Access confirmation show a warning when third-party libraries may inherit Secret Gate authority. Launcher identity remains exact designated-requirement equality; no path, display name, or Team ID alone gains authority.

ADR 0009 records the authority-boundary change.

## Validation

- `swift test --package-path src/menu-helper --disable-automatic-resolution` (76 tests)
- `swift build --package-path src/menu-helper --disable-automatic-resolution -c release`
- release `AutomicVaultMenubar --self-check-standalone-launchers`
- `cargo test --all-targets --all-features --locked` (824 unit tests plus integration targets)
- `git diff --check`

Refs #141
Follow-up to #102
Companion to #143

### DIFF
--- a/docs/adr/0009-support-library-validation-launchers.md
+++ b/docs/adr/0009-support-library-validation-launchers.md
@@ -1,0 +1,74 @@
+# ADR 0009: Support Launchers That Disable Library Validation
+
+- Status: Accepted
+- Date: 2026-08-09
+
+## Context
+
+ADR-backed enforcement introduced by pull request #102 requires Hardened
+Runtime for new Secret Gate Launcher rules. It permits JIT executable-memory
+exceptions but rejects disabled library validation. Claude Code's official
+Developer ID-signed executable enables Hardened Runtime together with JIT,
+unsigned executable memory, and disabled library validation. As a result,
+making its version-numbered executable selectable for issue #141 would still
+leave an important agent Launcher ineligible for ordinary gate policy.
+
+Apple describes disabled library validation as allowing a process to load
+third-party code that is not signed by Apple or the same Team. This weakens the
+Launcher's in-process trust boundary, but it is distinct from the entitlement
+that permits DYLD environment variables to inject code or change library search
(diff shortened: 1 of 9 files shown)


mxcl:
Folded into #142 as requested.

PR #145, opened 2026-08-09
fix(scan): dedupe bash/zsh PATH findings for the same directory

nbost130:
`av scan` runs separate bash and zsh detectors against the same `$PATH`, so every insecure directory is reported twice — once per shell, identical text otherwise. Anyone with both shells configured (the macOS default, plus anyone with a leftover `.bashrc`) sees roughly double the real finding count with no corresponding increase in risk.

This adds a merge pass after detectors run: bash/zsh PATH findings for the same directory collapse into one `bash+zsh`-sourced finding naming both shells. Other bash/zsh findings (e.g. plaintext credentials in startup files) and the separate macOS GUI-PATH detector are untouched.

Verified against a real scan: 42 duplicate PATH findings across 21 directories collapsed to 21 merged findings.

**Tests:** 5 new — the merge case, and the cases that must not merge (single-shell finding, different directories, non-PATH bash/zsh findings, the macOS GUI-PATH detector). `cargo test --lib`: 829 passed. `cargo fmt --check`: clean.

### DIFF
--- a/src/isotopes/detectors/mod.rs
+++ b/src/isotopes/detectors/mod.rs
@@ -395,12 +395,13 @@ const MERGED_SHELL_PATH_SOLUTION: &str = "Shell startup files contain arbitrary 
 fn merge_duplicate_shell_path_findings(findings: Vec<Finding>) -> Vec<Finding> {
     let mut merged: Vec<Finding> = Vec::with_capacity(findings.len());
     for finding in findings {
-        let existing = is_shell_path_finding(&finding).then(|| {
+        let source = finding.source;
+        let existing = shell_path_entry(&finding).and_then(|entry| {
             merged.iter_mut().find(|candidate| {
-                is_shell_path_finding(candidate) && candidate.affected == finding.affected
+                candidate.source != source && shell_path_entry(candidate) == Some(entry)
             })
         });
-        match existing.flatten() {
+        match existing {
             Some(existing) => merge_shell_path_finding(existing),
             None => merged.push(finding),
         }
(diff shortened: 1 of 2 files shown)


mxcl (changes requested):
The merge can suppress real PATH Findings rather than only deduplicating them.

`radioisotope::affected()` records only paths beginning with `/` or `~`, while `path_security` deliberately reports relative entries and `.` for empty PATH components. Those Findings all have `affected == []`, so `candidate.affected == finding.affected` merges different entries—including entries from the same shell—and can hide a real Hazard.

Reproduced with `PATH="first:second:/usr/bin:/bin"`: the result contains two `bash+zsh` Findings both naming `first`; `second` disappears. Please key the merge by the exact PATH entry, require the opposite shell source, and add relative/empty-entry regression tests.

nbost130:
Fixed. radioisotope::affected only records entries starting with / or ~, so every relative and empty PATH entry had affected == [] and they all compared equal. That let the merge collapse unrelated entries into one, and let two findings from the same shell merge with each other.

The merge now keys on the PATH entry parsed out of the explanation, which is present for every entry, and requires the two sources to differ. merge_shell_path_finding reads the entry from the explanation too, so merged output names the entry for relative and empty entries instead of leaving the original single shell text in place. That is the display half Copilot flagged.
 
The test helper built a non-empty affected for every path, so no test ever exercised a relative entry. It now calls radioisotope::affected rather than copying the rule, which meant widening that function from private to pub(super). That is the only change outside the file under review: one keyword, no behavior change, visibility extended only to the detectors module. If you prefer we can move that into another issue/PR. 
 
New regression tests cover your first:second repro, the empty entry reported as ., and two findings from the same shell that must not merge. 832 tests pass and cargo fmt --check is clean.
 
Separately, and not in this PR: SHELL_PATH_FINDING_MARKER has to stay textually identical to the format string in src/cli/shell_secrets.rs. If that wording ever changes, merging silently stops and findings quietly double again. A shared constant would close that, but it touches a file outside this fix, so I left it.

PR #147, opened 2026-08-10
refactor(scan): define the user-writable PATH reason once

### LINKED ISSUE #146: Shell PATH merge depends on a sentence duplicated across three files, with nothing enforcing the match
Shell PATH finding merging depends on a sentence that lives in another file, and nothing enforces the match.

`merge_duplicate_shell_path_findings` decides whether a finding is a shell PATH finding, and pulls the PATH entry out of it, by matching against a copy of the sentence text:

```rust
// src/isotopes/detectors/mod.rs:386
const SHELL_PATH_FINDING_MARKER: &str =
    "PATH has a user-writable directory before protected system directories";
```

The sentence itself is built somewhere else:

```rust
// src/cli/shell_secrets.rs:44
"{shell} PATH has a user-writable directory before protected system directories: {}",
```

If that wording is ever reworded, `shell_path_entry` returns `None` for every shell PATH finding, no merging happens, and every insecure directory silently goes back to being reported twice. There is no panic and no failing test, so the regression would only show up as users noticing doubled findings again.

The tests do not catch it either. The test helper builds its own copy of the sentence rather than going through the real reason-building code, so the two strings can drift apart without anything turning red.

Same sentence also appears a third time in `src/isotopes/detectors/macos/mod.rs:27`, with a different prefix.

Suggested fix: put the shared part of the sentence in one constant and have all three sites format from it, so a reword cannot desynchronise them. An integration test that runs real reason output through the merge would also close it, though the constant seems simpler.

Not urgent, and deliberately left out of #145 because it touches files outside that fix. Filing so it is written down.

nbost130:
Fixes #146.

The user-writable PATH reason was duplicated between the Bash/Zsh producer and the macOS GUI-PATH producer. PR #145 added a consumer that matched another copy of the same text. Rewording one copy could silently stop shell finding deduplication.

This branch puts the singular reason builder and its shared recognition text in `path_security`. Both producers call the builder, and the detector merge introduced by #145 matches `USER_WRITABLE_PATH_REASON`. The merged display sentence remains explicit because its plural grammar is intentionally different: “Bash and zsh PATH have…”.

Output is unchanged.

Validation:

- `cargo test --lib`: 832 passed
- `cargo fmt --check`: clean
- `git diff --check`: clean

### DIFF
--- a/src/cli/shell_secrets.rs
+++ b/src/cli/shell_secrets.rs
@@ -1,5 +1,7 @@
 use std::path::{Path, PathBuf};
 
+use crate::path_security::USER_WRITABLE_PATH_REASON;
+
 pub(crate) fn bash_reasons() -> Result<Vec<String>, String> {
     let home = home_dir()?;
     let mut paths = vec![
@@ -39,12 +41,7 @@ fn path_reasons(shell: &str) -> Vec<String> {
     let path = std::env::var_os("PATH").unwrap_or_default();
     crate::path_security::user_writable_entries_before_system_paths(&path)
         .into_iter()
-        .map(|entry| {
-            format!(
-                "{shell} PATH has a user-writable directory before protected system directories: {}",
-                entry.display()
-            )
-        })
+        .map(|entry| format!("{shell} {USER_WRITABLE_PATH_REASON}: {}", entry.display()))
         .collect()
 }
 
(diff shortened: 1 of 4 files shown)


nbost130:
Pushed a follow-up: sharing only the middle of the sentence still left each producer hand-writing its own prefix and `": "` separator, which is the same drift narrowed rather than removed. The separator matters too, since `radioisotope::affected` recovers the reported path by splitting on `": "`.

`path_security` now owns a small builder for the whole reason and both producers call it. The constant stays for consumers that recognise these reasons by matching the wording. Output unchanged, 824 tests pass.

PR #151, opened 2026-08-10
Add destination-gated Secret Proxy sessions

### LINKED ISSUE #127: Egress proxy: hand the child a placeholder, keep the secret in the vault
Hey @mxcl - I've been building a cli tool for a similar problem over at [than/secure-api-mcp](https://github.com/than/secure-api-mcp), and I'm excited to use `av`! Here's one thing I was looking at that's just outside what `av` currently covers.

## Where custody ends

What I like about av is that it's a custodian, not just a store. The secret sits encrypted, the app is the only thing that hands it out, and it decides case by case whether to. The dashboard can't even display a value.

Custody ends at the handoff. `av inject` decides *whether* to release a secret, puts the plaintext in the child's environment, and execs away. The manual is upfront that this is a handoff boundary and not a sandbox.

That's a reasonable boundary against a launcher that shouldn't have been trusted. It's weaker against one that *is* trusted but is being driven by an agent, because the approval is about identity and the risk is about destination. Approve `curl` once and whatever drives it picks where the key goes.

Coding agents also write everything down. Claude Code stores a full transcript of every session under `~/.claude/projects/`, including the stdout of every command it runs. Codex does the same under `~/.codex/sessions/`. I checked my own machine: 544 MB across 1,459 files, oldest 47 days back, much of it mode 0644, none of it behind a TCC prompt. Any `npm postinstall` running as me can read all of it. Sweeping for credential shapes turned up live-looking API tokens and JWTs.

One caveat that shapes what I'm claiming. Of 28 credential-shaped hits, 1 was in a command someone typed. The other 27 arrived as command *output* or file contents. So this idea doesn't touch the majority of what lands in those files — `av scan` and hardened tools do. It only helps with secrets on their way out.

## Proposal

Put a stable placeholder in the child's environment instead of the real secret.

```
av inject --proxy +GITHUB_TOKEN -- node app.js
```

- The child gets `GITHUB_TOKEN=av-ph-GITHUB_TOKEN-Qx9fLm2VtR8w` and `HTTPS_PROXY=http://av:<token>@127.0.0.1:<port>`.
- The app hosts a loopback proxy. It finds the placeholder in outbound requests, checks the destination against a per-secret allowlist, and substitutes the real value only for allowed hosts.
- Anything else raises the existing approval prompt: *"node app.js wants to send GITHUB_TOKEN to telemetry.evil.example."* Deny / Once / Session / Always.
- Each substitution writes an audit record through the existing access log.

Three things that identity gating can't give you:

1. Sending a secret somewhere unexpected becomes its own decision, not a side effect of an approval already granted.
2. The child never holds the secret. `env`, crash dumps, `ps`, logs, and agent transcripts all get an inert string.
3. Audit records carry a destination, not just "granted at 14:02".

## The obvious objection

A loopback port that turns placeholders into real secrets is new attack surface, and it's the part I'd most want your read on. The session token has to reach the child somehow, and wherever it lands, something else running as me can probably get at it. I couldn't find a way to attest the peer on a TCP loopback socket the way you can over a unix socket, and `HTTP_PROXY` wants a host:port. If I'm missing a trick there, it changes the design.

Today `av inject` puts the real secret in that same environment, where the same attacker gets it for free. After this, the best they do is replay it to a host I already approved and leave an audit record doing it. The policy does the real work; the token only keeps unrelated local software from stumbling into the port.

## What it doesn't cover

**Output redaction.** I looked at this and don't think it belongs here. `av inject` uses `exec()`, so there's no child, no pipes, no stdout to own. Redacting output means spawning instead, with `forkpty` and signal forwarding, and you lose the transparency `exec()` gives for free.

**Tools that transform the credential.** Basic auth base64s `user:pass`; SigV4 signs with the key. Those transform the placeholder, the proxy doesn't recognize the result, and the request fails upstream. It fails closed, but it's a real limit.

## What would break

Substituting into HTTPS means intercepting TLS: a session CA, per-host certs minted on the fly, and trust scoped through the environment (`SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, and friends) rather than the system trust store.

Plenty of tools ignore those variables — anything using Security.framework directly, and as I understand it a lot of Go binaries. Those fail closed rather than leak: they still route through the proxy, reject a cert they don't trust, and stop before sending anything. But proxy mode would cover a subset of tools on day one, and `av doctor` would need to say which ones before someone hits it mid-task.

You'd have a much better instinct than me for how big that subset is, and whether it's big enough to be worth it.

Whatever shape it took, it would want to fail closed like the rest of av: an empty allowlist allows nothing, an unreadable policy allows nothing, and the real host's certificate still gets verified on the way out.

Mostly I want to know whether this is a direction you'd want at all, or something you'd rather solve another way. 

Thanks!

mxcl:
## Summary

- add `av proxy +SECRET -- COMMAND`, which gives the Target random Secret References instead of raw Secret values
- add manual Proxy Session and exact-origin Destination authorization, Active Proxy controls, and bounded Authorization History
- isolate HTTP/S interception in a separately signed, sandboxed Hardened Runtime helper with no Keychain authority
- document the bearer boundary and compatibility limits in ADR 0011 and the Secret Proxy guide

## Why

Issue #127 proposes an egress proxy that lets agentic tools exercise arbitrary secrets without putting reusable values in their environment. This implements that capability as a general Automic Vault feature while preserving manual authorization and the existing Secret custody boundary.

## Security model

The signed `av` Gate Client requests every session manually. The app binds the session across the one approved `exec` transition using PID version, start time, effective UID, audit session, canonical Target path, and live code identity where macOS provides one. It revalidates identity after approval and immediately before releasing Secret values.

The loopback Proxy Credential and Secret References remain bearer values; PID binding limits session lifetime and does not attest individual TCP flows. Session rules bind an exact canonical origin plus exact Secret Names and are memory-only. The helper rejects private or reserved destinations, ambiguous DNS, invalid upstream TLS, protocol upgrades, uninspectable or oversized bodies, and transactions over 30 seconds.

## Product impact

- new proxy-specific Approval copy and Hardened Runtime warning
- new Active Proxies sidebar with destination statistics and termination
- terminated sessions leave records in the existing global 50-entry Authorization History
- no Launcher blessing or durable Destination policy in this first version

## Validation

- `cargo test`
- `swift test --package-path src/menu-helper`
- signed sandbox-helper black-box CONNECT, authentication, and private-address test
- `scripts/build.sh`
- strict nested code-sign verification and entitlement inspection

Closes #127


mxcl:

than:
Awesome!

mxcl:
LMK if it works how you expected or if you want any mods. Essentially: this is your feature.

Edit: This is live in 3.15.0.

PR #155, opened 2026-08-12
Fix Brew shellenv and PATH probe recursion

### LINKED ISSUE #152: `av harden brew` 3.2.0 on macOS 26: empty shellenv, getcwd failures from home, and user service data chowned under var
I ran `av harden brew` on macOS 26.5 (Apple Silicon, Homebrew 6.0.16, AV 3.2.0) and hit three separate problems. Reporting here first as the docs ask. Happy to split into three issues if you prefer.

One thing matters for the first two: my home directory is `drwxr-x---`. That is the default on newer macOS installs, not something I hardened myself.

## 1. `brew shellenv` from the launcher prints nothing

```
$ env -i PATH="/usr/local/bin:/usr/bin:/bin" HOME=/Users/me USER=me /usr/local/bin/brew shellenv
$ echo $?
0
```

Empty output, exit 0. Same for `shellenv sh` and `shellenv zsh`. Your docs say to put `eval "$(/usr/local/bin/brew shellenv)"` in shell startup, and the hardener rewrites `.zprofile` to do exactly that. Result: new shells get no `/opt/homebrew/bin` in PATH and every formula binary disappears from the shell. My guess: the launcher clears the environment before running the real brew, so brew inspects the launcher's internal env (already complete) instead of the caller's, and the idempotency check decides there is nothing to print. Workaround: hardcode the exports in `.zprofile`.

## 2. brew fails from any directory inside `$HOME`

The launcher runs as `automic`, which cannot traverse a `drwxr-x---` home. So from any cwd under my home:

```
shell-init: error retrieving current directory: getcwd: cannot access parent directories: Permission denied
Error: $PWD must be set to run brew.
```

Since a shell is almost always somewhere under `$HOME`, this breaks brew for daily use. Workaround that matches what you already get on older macOS with 755 homes, but narrower: `chmod +a "user:automic allow search" ~`. Suggestion: detect this in preflight and either apply the ACL with consent or document it.

## 3. Ownership repair breaks brew services data under `/opt/homebrew/var`

I had `postgresql@17` running through `brew services` (runs as the login user, data in `/opt/homebrew/var/postgresql@17`). The harden chowned that directory to `automic:vault` mode 700. Connections started failing with `could not open file "global/pg_filenode.map": Permission denied`, and the service cannot start after a reboot. Service logs under `/opt/homebrew/var/log` got chowned too, so launchd jobs cannot reopen them on restart. Suggestion: exclude user service state under `var` from the repair, or chown it to the invoking user.

Happy to give more details or test fixes on this machine.

mxcl:
Fixes #152.

## What changed

- Give the authorized `brew shellenv` child a sanitized system-only PATH that does not already contain the Homebrew prefix, so Homebrew emits its shell exports.
- Keep `shellenv` behind the existing Homebrew Execution Gate; this does not add an authorization bypass.
- Reject `--` before authorization so command classification cannot be bypassed.
- Bump the hardened Brew stub marker to V19 so Doctor detects older stubs as stale.
- Stop the menu bar app from executing the user's login-shell startup files to discover PATH.
- Suppress PATH-order and PATH-resolution findings in menu-bar Doctor because the app cannot verify the user's shell PATH safely. Non-PATH findings remain visible; `av doctor` continues to perform full PATH validation when run from a login shell.
- Report Doctor execution or decoding failures explicitly instead of mislabeling them as PATH failures.

## Upgrade note

Installing the updated app does not replace an existing root-owned setuid Brew launcher. Existing hardened installations must run `av harden brew` once to install the bundled V19 launcher.

## Root cause

The stub always launched Homebrew with `/opt/homebrew/bin:/opt/homebrew/sbin` first in PATH. Homebrew's `shellenv` treats that as already configured and returns successfully without output.

Separately, the menu bar app ran `$SHELL -lic` to inspect PATH. A normal `.zprofile` can call the hardened Brew stub, which synchronously asks the same app for authorization while the app is waiting for the login shell.

## Validation

- `cargo test --all-targets --all-features --locked`
- `swift test --package-path src/menu-helper --disable-automatic-resolution`
- Regression tests verify the system-only `shellenv` PATH, reject `--` before command classification, suppress PATH-derived menu-bar Doctor issues, and report Doctor load failures directly.

### DIFF
--- a/src/brew_stub/main.rs
+++ b/src/brew_stub/main.rs
@@ -6,9 +6,10 @@ use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-const MARKER: &str = "AUTOMIC_VAULT_BREW_STUB_V17";
+const MARKER: &str = "AUTOMIC_VAULT_BREW_STUB_V18";
 const TARGET: &str = "/opt/homebrew/bin/brew";
 const PREFIX: &str = "/opt/homebrew";
+const SHELLENV_PATH: &str = "/usr/bin:/bin:/usr/sbin:/sbin";
 const APPROVAL_SERVICE: &str = "com.automicvault.av2.approval";
 const BREW_USER_UID: &str = "/opt/homebrew/var/automic/user-uid";
 const FORBIDDEN_CASK_ARTIFACTS: &str = "app appimage artifact audiounitplugin bashcompletion colorpicker commandwrapper dictionary fishcompletion font generatedscript inputmethod installer internetplugin keyboardlayout manpage mdimporter pkg postflight postflightblock postflightsteps preflight preflightblock preflightsteps prefpane qlplugin screensaver service stageonly suite uninstall uninstallpostflightsteps uninstallpreflightsteps vst3plugin vstplugin zshcompletion";
@@ -69,6 +70,8 @@ where
     F: FnOnce(&AuthorizationRequest) -> Result<(), String>,
 {
     let mut request = authorization_request(&args, cwd)?;
+    let shellenv =
+        command_index(&request.args).is_some_and(|index| request.args[index] == "shellenv");
     let (args, cask) = governed_args(&request.args)?;
     request.args = args;
     if let Some(cask) = &cask {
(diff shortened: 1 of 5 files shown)


mxcl:

PR #164, opened 2026-08-14
Harden Docker registry credentials

### LINKED ISSUE #161: Docker credential-helpers finding: no approval-aware hardener exists yet for credsStore desktop
## Summary

The `docker-credential-helpers` detector flags: "Docker config uses
ambient credential store `desktop`." Per the detector's own docs,
"Rewriting the existing helper configuration is not enough" — a real fix
needs "an approval-aware helper that preserves Docker's credential-helper
protocol and registry routing," which doesn't exist today.

## What I found on this machine

- `~/.docker/config.json` sets `"credsStore": "desktop"`, delegating all
  registry credential storage to `docker-credential-desktop` — a plain,
  ungated binary any process running as the user can invoke.
- Nothing is actually stored there right now (`docker-credential-desktop
  list` → `{}`), so there's no live exposure today — but the ambient
  mechanism means the next `docker login` would be exposed the same way
  the git credential-helper finding was, until this is fixed.

## Question

Is there a roadmap for an approval-gated `docker-credential-vault`-style
helper — implementing Docker's credential-helper protocol (`get`/`store`
/`erase`/`list` over stdin/stdout) — that routes through the same "Secret
Gate" approval flow `av harden gh` already provides for git? That would
let `credsStore` point at a signed Automic Vault binary the same way
`credential.helper` now points at the signed `gh` Isotope.

Happy to help test/validate against this machine's setup if useful.

mxcl:
Closes #161.

## What changed

- add `av harden docker` and a root-owned `docker-credential-av` launcher
- migrate credentials from Docker Desktop's `desktop` or `osxkeychain` helper into Automic Vault with save-before-delete semantics and rollback
- keep Docker Desktop's vendor-signed Docker, Compose, and Buildx CLIs instead of producing isotopes
- verify Docker's notarization, Developer ID team, signing identifiers, Hardened Runtime, timestamp, and unsafe-entitlement absence
- authorize helper reads, stores, and erases through XPC using the live Docker parent, exact PID start time, UID, executable, arguments, code identity, registry, and derived Secret Name
- reject inline credentials, non-Automic per-registry helpers, malformed protocol input, and helper `list` metadata disclosure
- document the resulting security boundary in ADR 0014

## Why

Docker's credential-helper protocol protects secrets at rest but returns a usable registry credential as plaintext JSON to any caller of `get`. The hardener moves custody and authorization into Automic Vault while retaining Docker's stronger vendor distribution identity. The plaintext handoff remains only at the unavoidable protocol boundary, after the actual Docker Target has been authorized and revalidated.

## Validation

- `cargo test --all-targets` — passed
- `cargo test docker --lib` — passed
- `swift test` — 100 tests passed
- `AutomicVaultMenubar --self-check-docker-credentials` — passed
- `AutomicVaultMenubar --self-check-secret-mutations` — passed
- hardener metadata successfully verified the installed Docker Desktop signatures and runtime posture

`cargo clippy --all-targets -- -D warnings` remains blocked by the repository's existing Rust 1.96 lint backlog outside this change.

### DIFF
--- a/docs/adr/0014-docker-credential-helper.md
+++ b/docs/adr/0014-docker-credential-helper.md
@@ -21,8 +21,10 @@ establish Target integrity.
 
 The Docker Hardener installs an exact root-owned
 `/usr/local/bin/docker-credential-av` launcher for the signed Automic Vault
-CLI and changes Docker's default credential store to `av`. It does not replace
-Docker's vendor CLI.
+CLI and changes Docker's default credential store to `av`. Every containing
+directory through the filesystem root must also be a real, root-owned directory
+protected from group/world writes; installation and detection fail closed if
+that invariant is not met. It does not replace Docker's vendor CLI.
 
 Each registry credential is stored as one opaque Secret. Its Secret Name is a
 fixed prefix plus the SHA-256 digest of the exact registry address. The stored
--- a/docs/adr/0014-docker-credential-helper.md
+++ b/docs/adr/0014-docker-credential-helper.md
@@ -1,0 +1,76 @@
+# ADR 0014: Bind Docker registry credentials to live vendor-signed Targets
+
+Status: accepted
+
+## Context
+
+Docker Desktop configures `credsStore` to invoke a credential helper. The
+helper protects registry credentials at rest, but Docker's protocol returns a
+usable password, personal access token, or identity token as plaintext JSON to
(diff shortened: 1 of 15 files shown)


mxcl:
@copilot do you consider your issue resolved with the latest commit?

mxcl:
@copilot do you consider your issue resolved with the last commit?

PR #165, opened 2026-08-14
Design Launcher Bundles for unsigned CLI executables

mxcl:
## Summary

- define **Launcher Bundle** as the canonical artifact for packaging one unsigned Mach-O CLI as a potential Verified Launcher
- record creation, signing, enrollment, replacement, runtime verification, and failure semantics in ADR 0013
- pin each generation to its exact signed code identity and final payload digest, with no ordinary-launcher fallback on mismatch
- replace the earlier “Secure Launcher” wording in architecture, provenance documentation, and the menu-helper documentation link

## Why

Unsigned and arbitrary ad-hoc-signed CLI executables do not provide a persistent Launcher Identity. Issue #137 needs a user-attended way to package a single CLI payload without introducing a new gate or overstating the resulting artifact security.

This PR establishes the domain and security boundary before implementation.

## Impact

This is primarily a design and terminology change. The Swift change only renames the existing documentation URL constant and link target; it does not implement Launcher Bundle creation yet.

## Validation

- `git diff --check origin/main...HEAD`
- `cd src/menu-helper && swift test` — 100 tests passed

Part of #137.


mxcl:

Issue #176, opened 2026-08-19
BUG: av inject stops working in 3.9.0


phihung:
I updated AV to 3.9.0 (both app and CLI) yesterday, and `av inject` has stopped working.

`av inject +AAA -- echo "hello"` hangs indefinitely and never prompts for permission.

At one point, AV took up to 25 GB of RAM, so I had to kill it. But I have not been able to reproduce the issue since.

My Mac is an M3 Max running `macOS 15.7.1`.

<img width="746" height="144" alt="Image" src="https://github.com/user-attachments/assets/69ed5d09-ad65-428a-bd96-5693bb5e8c69" />

mxcl:
Not great if generally true, I'm taking a look. Sorry for this issue. I will also spin up some stuff to look for how we managed to take up 25GB of RAM as a separate issue. I have not seen that, but I'm sure there is a way.

In the meantime downgrading should be safe and effective.

mxcl:
Thanks for reporting this. Initial triage:

- The CLI’s `av inject` implementation is unchanged from 3.8.0.
- The official 3.9.0 Approval self-check passes on another M3 Max, but that machine runs macOS 26.5.2 rather than 15.7.1, so this does not rule out a macOS 15-specific issue.
- I haven’t reproduced the hang or the 25 GB memory growth yet.

The most useful distinction is whether the terminal prints:

`automic vault: human approval required`

If this happens again, please leave it hanging long enough to capture the following before restarting Automic Vault:

```sh
pid=$(pgrep -x AutomicVaultMenubar)
ps -o pid,rss,vsz,etime,state,command -p "$pid"
sample "$pid" 10 -file ~/Desktop/AutomicVault-176.sample.txt
```

Please also say whether the Automic Vault main window remains responsive and whether an Approval window appears in Mission Control or another Space.

Review the sample before posting it publicly because stack samples can contain local filesystem paths.

I'll continue looking through potential changes for regressions, and more long term codepaths that may be responsible.

phihung:
Thank you for your quick reply
 
- The terminal does NOT print `automic vault: human approval required`
- Automic Vault main window remains responsive
- No Approval window appears in Mission Control or another Space
- No trace in Authorization history

One more detail I don't know if it helps: running `av inject -- echo "hello"` triggers the Approval window.

[AutomicVault-176.sample.txt](https://github.com/user-attachments/files/31226746/AutomicVault-176.sample.txt)

mxcl:
[...]

PR #177, opened 2026-08-19
Reply when Approval presentation is interrupted

mxcl:
Hardens a separate accepted-without-reply path found while auditing #176. The reporter's supplied sample identifies cyclic Project Directory ancestry as that issue's root cause; this PR remains an independent fix.

When macOS aborts the modal Approval because the user session resigns active or the displays sleep, the Gate Client is still connected. The prompt previously returned `.canceled`, which every XPC caller interprets as “the Gate Client exited” and therefore intentionally does not reply. The live CLI then waits indefinitely.

Preserve a distinct interrupted result and return a fail-closed system error to the live client. Actual XPC cancellation still returns `.canceled` and sends no reply because the client is gone. An interruption is recorded as a system failure, emits no human-decision field, and is never stored in the transient user-denial cache. Explicit prompt denial keeps its existing behavior.

Based directly on the `3.9.0` tag for inclusion in 3.9.1.

Verification:
- `swift test --package-path src/menu-helper --disable-automatic-resolution` (132 tests)
- `AutomicVaultMenubar --self-check-approvals`
- `AutomicVaultMenubar --self-check-transient-approvals`
- `AutomicVaultMenubar --self-check-secret-mutations`

### DIFF
--- a/src/menu-helper/Sources/MenubarHelper/main.swift
+++ b/src/menu-helper/Sources/MenubarHelper/main.swift
@@ -1886,6 +1886,7 @@ private struct TransientApprovalKey: Hashable {
 
 private enum ApprovalDecision: Equatable {
     case canceled
+    case interrupted
     case denied
     case approved
     case alwaysApproved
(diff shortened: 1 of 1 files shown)


mxcl:
@copilot do you consider your issue resolved with the latest patches?

PR #180, opened 2026-08-20
Connect approval relay only while requests are pending

mxcl:
## Summary

- connect each Mac to the Approval relay only while it has pending Approvals
- establish the receive path before publishing, keep it live with pings, and reconnect and republish on failure
- remove the iPhone misleading idle Connected Macs status

## Security

- preserves fail-closed behavior and Mac-side response validation
- leaves the relay protocol and authority model unchanged

## Testing

- swift test --package-path src/menu-helper --disable-automatic-resolution: 147 tests passed
- iPhone simulator Debug build succeeded
- deployed to Maliwan, Pangolin, and Anshin; verified both idle Macs maintain zero relay sockets


mxcl:

PR #194, opened 2026-08-24
Control temporary access grant countdowns

mxcl:
## Summary
- add an End-button dropdown with Add 10 Minutes and Pause or Resume Write Access actions
- extend running or paused countdowns without reviving expired grants
- disable grant authority while remaining active time is frozen
- show the enclosing app display name instead of its launcher path
- remove the duplicate dropdown chevron

## Security
Extensions are explicit and retain the same Tool-specific gate, Verified Launcher, runtime posture, and agent task scope. Pausing does not leave Write Access active, adding time to a paused grant does not resume it, and expired grants cannot be revived. Sleep, session, update, app-termination, and explicit-end revocations remain unchanged. ADR 0026 records the boundary change.

## Testing
- swift test --package-path src/menu-helper --disable-automatic-resolution (182 tests)
- AutomicVaultMenubar --self-check-standalone-launchers
- AutomicVaultMenubar --self-check-menu-status
- manually verified a GH local-write grant and Add 10 Minutes using gh repo clone mxcl/AppUpdater

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,7 +339,7 @@ Launcher, Tool-specific gate, runtime posture, and current agent task.
 
 <img src="./docs/img/temporary-write-access.png" alt="Automic Vault temporary write access controls" style="width: 589px; height: auto" />
 
-The persistent strip shows every grant and lets you add ten minutes, pause or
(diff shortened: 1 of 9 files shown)


mxcl:

PR #205, opened 2026-08-27
Document gh credential-resolution failure semantics

saltedzucchini:
Depends on automic-vault/gh-cli#3

## Summary

Document the fail-closed credential-resolution contract for the hardened `gh` Isotope.

The updated hardener guide distinguishes exact credential absence from denial, timeout, and approval-service or XPC unavailability. Operational failures must stop before a GitHub request. `gh auth status` must report `Vault retrieval unavailable.` instead of describing a valid account as logged out or invalid.

The guide also records the tested API, watcher, and refresh boundaries. It links the public upstream issue and pull request without repeating their stale timeout claim.

## Release boundary

The text describes the intended Hardened State. It does not claim that a checkout or currently installed binary already has the fix.

Only a verified official signed Automic Vault `gh` Isotope installed through `av harden gh` may use real credentials. Local source builds remain test-only. Environment tokens, plaintext credential files, and direct-binary workarounds are not equivalent hardened routes.

## Validation

- `git diff --check`
- `cargo +1.96.0 fmt --all -- --check`
- Independent review against the canonical domain language, architecture, and positioning documents

The branch changes one Markdown file. It does not change product code, credentials, policy, or deployment state.

## Authorship and follow-up

An agent wrote this under close human direction. An agent may draft review replies, but `@saltedzucchini` will review them before posting.

### DIFF
--- a/src/isotopes/hardeners/gh_cli.md
+++ b/src/isotopes/hardeners/gh_cli.md
@@ -5,8 +5,9 @@
 We provide a [patched version] of `gh`. `av harden gh` installs it from our
 [tap] when Homebrew is available, or installs the same signed release directly
 at `/usr/local/bin/gh`. The supported hardened path is designed so that only
-the verified `gh` Isotope can apply its GitHub Secret through the Automic Vault
-Secret Gate.
+the verified `gh` Isotope submits authenticated operations through the `gh`
+Secret Gate; Automic Vault applies the GitHub Secret only to the authorized
+Target.
 
 The statements below describe the intended hardened-state contract. They do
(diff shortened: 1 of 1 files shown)


mxcl:
Closing as superseded by automic-vault/gh-cli#4. The shipped fix preserves credential-provider failures through the normal HTTP transport path, so this is an implementation correction rather than a new public contract. The canonical Automic Vault documentation already requires uncertainty to fail closed without falling back to broader access, and the regression tests now preserve the specific behavior. No additional hardener documentation is needed.

PR #238, opened 2026-09-01
Fix undismissable Direct Access confirmation sheet

### LINKED ISSUE #235: 3.18.2: "Allow Launcher…" presents an empty sheet that cannot be dismissed and locks the main window
## Summary

Adding a Direct Secret Access Launcher opens a sheet with no content. The sheet
has no buttons, does not close, and blocks the main window for the life of the
process. The app itself stays responsive — its main thread sits idle in the
normal event loop — so this is a stuck presentation, not a hang.

The sheet's content is built from a second `@State` value that is nil when the
sheet is presented, and the empty branch renders nothing at all.

## Steps to reproduce

1. Select a Secret in the main window.
2. Under **Direct Secret Access**, click **Allow Launcher…**.
3. In the open panel, choose `/Applications/Visual Studio Code.app` and click Choose.

Result: an empty sheet appears over the main window. Nothing in the window can be
clicked. Closing the window and reopening it from the menu bar brings the same
empty sheet back on top. Only quitting the process clears it.

## Expected

Either the `DirectAccessConfirmationView` with Allow/Cancel, or the
"cannot be allowed" alert if the Launcher is not eligible. Any sheet the app
presents should have a way out.

## The app is not hung

`sample` of the stuck process, twelve minutes after the sheet appeared, 0% CPU:

```
4174 Thread_4382735   DispatchQueue_1: com.apple.main-thread  (serial)
  4174 start  (in dyld) + 6076
    4174 main  (in AutomicVaultMenubar) + 2832
      4174 -[NSApplication run]  (in AppKit) + 480
        4174 -[NSApplication(NSEventRouting) _nextEventMatchingEventMask:untilDate:inMode:dequeue:]
          4174 _DPSNextEvent  (in AppKit) + 684
            ...
              4174 mach_msg2_trap  (in libsystem_kernel.dylib) + 8

Binary Images:
0x104d28000 - 0x105337567 +com.automicvault (3.18.2 - 3.18.2) /Applications/Automic Vault.app/Contents/MacOS/AutomicVaultMenubar
```

No modal loop on the stack, and the status-bar menu still dispatches actions.

## What the window log shows

`log show --predicate 'processIdentifier == <pid> AND subsystem == "com.apple.AppKit"'`:

```
11:40:24.208  [AppKit:OpenSavePanels] Advance to run phase.
11:40:24.260  order window front conditionally: 2123            <- the open panel
11:40:33.214  order window: 2123 op: 0                          <- panel dismissed (Choose)
11:40:37.592  parent <MenubarHelper.AutomicVaultWindow ...> windowNumber=211e
              add <SwiftUI...SheetPresentationWindow: 0x1321e2c70> windowNumber=2128 to group
11:40:37.618  order window: 2128 op: 1 relative: 211e           <- sheet above the main window
```

and then, on every later attempt to open the app from the menu bar
(11:40:49, 11:41:52, 11:48:46, 11:49:09 — identical each time):

```
perform action for menu item
Window <MenubarHelper.AutomicVaultWindow ...> windowNumber=211e ordered front
  from a non-active application and may order beneath the active application's windows.
order window front conditionally: 211e
order window front conditionally: 2128
order window: 211e op: -1 relative: 2128                        <- main window back under the sheet
```

No `_NSAlertPanel` is created in this run, so `showLauncherCannotBeAllowed` was
not the path taken — the completion fired and the sheet was presented.

## Root cause

`src/menu-helper/Sources/MenubarHelper/MainWindow.swift` (line numbers at tag 3.18.2):

```swift
// 2578
Button {
    model.chooseDirectAccessLauncher { launcher in
        guard let launcher else { return }
        pendingDirectAccessLauncher = launcher
        isConfirmingDirectAccess = true
    }
} label: {
    Label("Allow Launcher…", systemImage: "app.badge.checkmark")
}

// 2664
.sheet(isPresented: $isConfirmingDirectAccess, onDismiss: {
    pendingDirectAccessLauncher = nil
}) {
    if let selection = pendingDirectAccessLauncher {
        DirectAccessConfirmationView(...) { ... }
    }
}
```

Presentation is driven by `isConfirmingDirectAccess` while the content is built
from `pendingDirectAccessLauncher`. When the content closure runs with that
value still nil, the `if let` produces an empty view — and because the only
dismiss control lives inside `DirectAccessConfirmationView`, the sheet has no
way to close. `onDismiss` never runs, so the state is never cleared and the
sheet is re-presented every time the window is reopened.

The two `@State` writes also happen from `panel.begin`'s completion handler,
outside any SwiftUI transaction.

## Suggested fix

Drive the sheet from the value itself, which cannot reach this state:

```swift
.sheet(item: $pendingDirectAccessLauncher) { selection in
    DirectAccessConfirmationView(...)
}
```

The same file already does exactly this two lines above, for `replacingValue`.

Independently: every sheet content branch should end in something dismissable,
so a nil binding degrades to a Cancel button rather than to a locked window.

## Related, in the same flow

`chooseDirectAccessLauncher` (MainWindow.swift:823) calls `completion` only on
success — cancelling the picker and the ineligible-Launcher path both `return`
without invoking it. Not the cause of this bug, but it means a caller cannot
distinguish "cancelled" from "rejected".

## Environment

- Automic Vault 3.18.2 (updated from 3.18.0 shortly before; the bug occurred after the update)
- macOS 15.7.2 (24G325), Apple silicon
- Chosen Launcher: Visual Studio Code 1.x, `com.microsoft.VSCode`,
  Team ID UBF8T346G9, Hardened Runtime (`flags=0x10000(runtime)`),
  entitlement `com.apple.security.cs.allow-jit`
- The Secret involved has a Global Value only

mxcl:
## Summary

- drive the Direct Access confirmation sheet from the selected Launcher
- remove the independent Boolean presentation state
- identify the ephemeral selection by its existing designated requirement

## Root cause

Commit `4f2fe168` split presentation across `isConfirmingDirectAccess` and `pendingDirectAccessLauncher`. SwiftUI could present from the Boolean while the optional item still rendered nil, producing an empty sheet with no dismiss control.

Using `.sheet(item:)` makes presentation and content share one state, so the empty-sheet state is unrepresentable.

## Security

This changes presentation state only. Launcher eligibility, runtime-protection checks, the Direct Access warning, Approval, Direct Access Rule persistence, and fail-closed behavior are unchanged.

## Verification

- deterministic source invariant: red before, green after
- `swift test --package-path src/menu-helper --disable-automatic-resolution` — 240 runnable tests passed; entitlement-dependent Keychain tests skipped as expected
- `scripts/test-menu-helper-self-checks.sh` — 26 passed
- `scripts/test-launcher-bundle-runner.sh` — passed
- `scripts/test-varlock-plugin-helper.sh` — passed

Closes #235

### DIFF
--- a/src/isotopes/hardeners/homebrew.rs
+++ b/src/isotopes/hardeners/homebrew.rs
@@ -24,7 +24,7 @@ const BREW_USER_UID_FILE: &str = "var/automic/user-uid";
 const LEGACY_CASK_USER_UID_FILE: &str = "var/automic/cask-user-uid";
 const STUB_MARKER_PREFIX: &[u8] = b"AUTOMIC_VAULT_BREW_STUB_V";
 #[cfg(test)]
-const STUB_MARKER: &[u8] = b"AUTOMIC_VAULT_BREW_STUB_V19";
+const STUB_MARKER: &[u8] = b"AUTOMIC_VAULT_BREW_STUB_V20";
 const STUB_VERSION: u32 = 20;
 const ID_RANGE: std::ops::RangeInclusive<u32> = 550..=599;
 
@@ -1539,7 +1539,7 @@ mod tests {
         assert!(is_managed_stub_file(&path));
         assert!(!stub_is_current(&path));
 
-        fs::write(&path, b"AUTOMIC_VAULT_BREW_STUB_V20 future").unwrap();
+        fs::write(&path, b"AUTOMIC_VAULT_BREW_STUB_V21 future").unwrap();
         assert!(stub_is_current(&path));
 
         for invalid in [
--- a/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
+++ b/src/menu-helper/Sources/MenubarHelper/MainWindow.swift
@@ -2680,7 +2680,6 @@ private struct StoredSecretDetailView: View {
     let secret: StoredSecret
     @State private var isAvailableWhileLocked: Bool
     @State private var isConfirmingDelete = false
-    @State private var isConfirmingDirectAccess = false
     @State private var pendingDirectAccessLauncher: DirectAccessLauncherSelection?
     @State private var replacingValue: StoredSecretValue?
     @State private var deletingValue: StoredSecretValue?
@@ -2774,7 +2773,6 @@ private struct StoredSecretDetailView: View {
                     model.chooseDirectAccessLauncher { launcher in
                         guard let launcher else { return }
                         pendingDirectAccessLauncher = launcher
-                        isConfirmingDirectAccess = true
                     }
                 } label: {
                     Label("Allow Launcher…", systemImage: "app.badge.checkmark")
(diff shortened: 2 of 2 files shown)


mxcl:
Follow-up: with authorization to fix the mainline Rust failure, commit a928e66f aligns the test-only Homebrew stub marker with production V20 and advances the forward-compatibility fixture to V21. The focused regression, full debug Rust CI profile, and full optimized Rust CI profile all pass locally.

PR #241, opened 2026-09-01
Fix Doctor launcher path collision guidance

byronrode:
## Summary
- detect when an executable occupies an environment wrapper’s reserved Launcher path but no separate Target exists
- report one precise `target_at_launcher_path` issue instead of contradictory missing-target and invalid-stub repairs
- preserve the fail-closed hardener behavior by directing the executable to be reviewed and preserved before the Launcher is installed
- add an end-to-end Doctor regression test for the `sentry-cli` layout

## Verification
- `cargo test --all-targets --all-features --locked` (1,036 tests passed)
- `git diff --check`

### DIFF
--- a/src/cli/doctor.rs
+++ b/src/cli/doctor.rs
@@ -565,6 +565,8 @@ fn diagnose_command(hardener: &str, command: &HardenerCommand, path: &OsStr) -> 
 }
 
 fn executable_at_reserved_launcher_path(command: &HardenerCommand) -> Option<&str> {
+    // Environment wrappers retain the bare Command name only when they could
+    // not resolve a separate Target outside the reserved Launcher path.
     (Path::new(&command.target_path) == Path::new(&command.name))
         .then(|| command.stub_path.as_deref())
         .flatten()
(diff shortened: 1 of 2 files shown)


mxcl:
we think this is a deeper problem.

Out of interest who is responsible for you sentry-cli install to /usr/local/bin?

mxcl:
@byronrode yo

mxcl:
One option is ofc allowing `/opt/av/bin` as a PATH.

I am reluctant however, as GUI apps will *never* (almost never) add this PATH. Which means a whole category of issues for AV users.

I will review this PR. It may be the correct PATH, ie: aborting hardening if something is already there.

mxcl:
The Hardener already fails closed for this layout: it excludes the reserved Launcher path from Target discovery, then reports `sentry-cli is not installed on PATH`. The refusal is correct; the message is misleading when the executable is present at `/usr/local/bin/sentry-cli`.

The actionable explanation belongs at that Hardener failure, so users do not need to run Doctor to understand why hardening failed. We will replace this PR with a shared environment-wrapper Hardener fix that identifies the occupied path and explains how to preserve and relocate or reinstall the Target before retrying. Hardening will continue to abort before installation or credential migration; it will not overwrite the existing executable.

Closing this PR in favor of that focused fix. Thanks for surfacing the confusing failure.

mxcl:
Replacement Hardener fix: #336. The clearer failure applies across environment wrappers; all 62 focused tests pass.

PR #243, opened 2026-09-01
Fix stale scans for unattributed detector findings

### LINKED ISSUE #231: Menu bar dashboard never re-scans detectors whose findings are unattributed (no watched file)
tl;dr: the dashboard only re-checks a detector when a file it's watching
changes on disk. This particular detector's live check doesn't have a file
to point at, so its tile in the dashboard can go stale in either direction
(look clean when it isn't, or look flagged after you've actually fixed it)
until something unrelated happens to trigger a full re-scan.

The menu bar app doesn't poll on a timer. It re-runs `av scan --json` only when a
watched file changes: `startDetectorWatchers` (`src/menu-helper/Sources/MenubarHelper/main.swift`) builds its watch
list from each detector's declared `watchScopes` plus the `affected` paths of
whatever findings the last scan produced.

That breaks down for any detector whose finding has no affected file. Right
now that's `git-credential-fill`'s live-probe path: it deliberately returns
`high_unattributed` (`src/isotopes/detectors/git/credential_fill.rs`, the
helper itself is defined in `src/isotopes/detectors/git/config.rs`) because
Git doesn't report which helper answered a `git credential fill` request.
There's no file to add to the watch list, so nothing about that detector ever
triggers a re-scan on its own.

I hit this directly: I had a stale `osxkeychain`-cached GitHub credential
answering `git credential fill` underneath the signed `gh` Isotope. I deleted
it from Keychain Access, confirmed with the CLI that the probe now goes
through the isotope's approval gate instead of answering silently, but the
dashboard still showed `git credential oauth` / `fill` / `credentials file` as
"Detector only." (its clean-state label, `detectorItems` in
`src/menu-helper/Sources/MenubarHelper/MainWindow.swift`) from before I made the change. Deleting a Keychain item
isn't a filesystem write to anything in `watchScopes`, so the watcher had no
signal to act on.

It'd have gone unnoticed in the other direction too: if the leak reappeared
(say the credential got re-cached the way it originally got there), the
dashboard would just keep showing "Detector only." until something unrelated
happened to trigger a full re-scan.

The workaround is a manual refresh (the toolbar's circular-arrow button, or
relaunching the app, since `scheduleScan(after: 0)` runs on launch), so
nothing's actually broken, it's just silently stale, with no indication in the
UI that "Detector only." might be stale rather than current.

Two directions that seem reasonable, not tied to either:

1. Give detectors capable of producing an unattributed finding a time-based
   fallback re-scan (every few minutes) alongside the file watch, since by
   definition there's nothing to watch for them.
2. Show a "last checked" timestamp on the tile, so a stale "Detector only."
   at least looks stale instead of looking current.

Filing this separately from #230 since the fix lives in the Swift
watch-scheduling code, not the Rust config parser that issue is about, even
though both trace back to the same root cause: this detector's findings can't
be attributed to a file.

victorkardel:
Fixes #231

## Cause and fix

The menu helper only scheduled detector rescans from declared watch scopes and affected finding paths. `git-credential-fill` can return a live finding without an affected path, so its state could remain stale when Keychain-backed credentials changed. SIP also reports unattributed live state.

Detector metadata now explicitly identifies detectors that can emit unattributed findings. The menu helper keeps its scoped file watchers and adds one five-minute timer that re-scans only that metadata-defined set. The timer is not reset by normal watcher rebuilds, preventing unrelated filesystem activity from indefinitely delaying the fallback scan.

## Security reasoning

This only adds detection refreshes; it does not alter authorization, findings, remediation, or fail-closed behavior. The detector set is explicit and defaults to false for older CLI metadata.

## Tests

- `swift test --package-path src/menu-helper`
- `swift run --package-path src/menu-helper AutomicVaultMenubar --self-check-scan-scheduling`
- Rust tests were not runnable locally because Cargo is not installed in this environment.

## Rollback

Revert this PR to remove the metadata field and fallback timer; existing file-watch scheduling remains unchanged.

### DIFF
--- a/src/cli/scan.rs
+++ b/src/cli/scan.rs
@@ -69,6 +69,7 @@ pub(crate) fn run_detectors_json<W: Write>(stdout: &mut W) -> i32 {
                     "path": scope.path,
                     "recursive": scope.recursive,
                 })).collect::<Vec<_>>(),
+                "can_produce_unattributed_findings": detector.can_produce_unattributed_findings,
             })
         }).collect::<Vec<_>>(),
     });
@@ -544,6 +545,7 @@ mod tests {
         assert!(output.contains(r#""name":"git-credentials-file""#));
         assert!(output.contains(r##""documentation":"# git-credential-fill Detector"##));
         assert!(output.contains(r#""watch_scopes":[{"path":"#));
+        assert!(output.contains(r#""can_produce_unattributed_findings":true"#));
     }
 
     #[test]
--- a/src/isotopes/detectors/mod.rs
+++ b/src/isotopes/detectors/mod.rs
@@ -161,6 +161,7 @@ pub(crate) struct DetectorMetadata {
     pub(crate) docs_url: String,
     pub(crate) documentation: &'static str,
     pub(crate) watch_scopes: Vec<DetectorWatchScope>,
+    pub(crate) can_produce_unattributed_findings: bool,
 }
 
 pub(crate) struct DetectorWatchScope {
@@ -178,6 +179,7 @@ struct Detector {
     findings: fn(&Path) -> Vec<Finding>,
     docs_url: &'static str,
     documentation: &'static str,
+    can_produce_unattributed_findings: bool,
 }
 
 #[cfg(test)]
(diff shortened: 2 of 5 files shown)


mxcl:
Related #231.

I haven't fixed it yet because periodic rescans are not super great IMO. But I haven't got a better idea bar a rethink on UX.

PR #246, opened 2026-09-02
Harden Fastly CLI static tokens

mxcl:
## Summary

- add a Fastly CLI Hardener that migrates named static tokens into Secret Custody and leaves only `@av` markers
- bind retrieval and mutation to the signed live Fastly Target, exact arguments, token name, and official API endpoint
- install the signed `fastly-cli-isotope` release through the Isotopes tap, with direct single-binary fallback when Homebrew is unavailable
- fail closed for SSO, legacy profiles, alternate endpoints, and unknown auth fields
- document the boundary in ADR 0039 and retire the report-only detector guidance

Isotope: https://github.com/automic-vault/fastly-cli/releases/tag/v16.0.0

## Verification

- `cargo test -- --test-threads=1`
- `swift test` in `src/menu-helper`
- `AutomicVaultMenubar --self-check-fastly-credentials`
- full upstream Fastly CLI Go test suite, plus release-manifest tests
- downloaded release asset passes strict code-signature verification with identifier `fastly`, team `ZU76A67LGU`, Hardened Runtime, timestamp, and no entitlements


mxcl:
@copilot review

mxcl:
The finding was valid, but I revised the proposed fix in 999d68b1. Current Fastly CLI selects XDG or macOS Application Support via `os.UserConfigDir`; it does not prefer an existing `~/.fastly` file. Treating that legacy file as authoritative would harden one config while the signed Target reads another. The hardener now fails closed when any inactive/legacy config exists, names both paths, and requires the user to move or merge it into Fastly’s active path. The detector guidance and regression test cover this boundary.

mxcl:
@copilot review

PR #253, opened 2026-09-02
Skip AWS approval for credential-free metadata

mxcl:
## Summary

- run exact AWS version and help invocations without making a Secret Use request
- keep target verification, environment scrubbing, and fail-closed handling for all other arguments
- use a fixed system path for help rendering and clarify Approval Required documentation

## Testing

- cargo test --all-targets --all-features --locked
- cargo test --profile test-release --all-targets --all-features --locked
- exercised aws --version and aws help through the launcher with an empty environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ Verified Launcher.
 
 Access Levels include:
 
-1. **Approval Required:** Automic Vault asks for Approval on every operation.
+1. **Approval Required:** Automic Vault asks for Approval on every Secret Use or
+   gated execution request.
 2. **Read Only:** Automic Vault automically authorizes recognized reads.
 3. **Read & Update:** Automic Vault automically authorizes Homebrew reads and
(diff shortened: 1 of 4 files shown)


mxcl:
Security review complete. No findings.

- The generation-bound stub and verified official AWS target checks still run before the metadata branch.
- The bypass accepts only exact version/help grammar; options, malformed arguments, and ordinary operations remain on the existing Secret Use path.
- Metadata runs with HOME=/var/empty, both AWS config sources disabled, instance metadata disabled, ambient AWS variables removed, and a fixed root-owned system PATH used only for help rendering.
- No Secret is selected, released, or recorded because this path makes no Secret Use request, matching the existing Hardener boundary in docs/domain-language.md.
- Debug and optimized test suites, the launcher-level regression test, and live credential-free version/help checks pass.

PR #257, opened 2026-09-03
Add a decision guide for which authorization mechanism to use

technicalpickles:
## Summary

- Add `docs/choosing-a-mechanism.md`, a "your situation -> mechanism" table covering Detectors, Hardened Tools, Authorization Gates, Blessed Scripts, Launcher Bundles, the Secret Proxy, Direct Secret Access, Launcher Bundles, and Authorization History
- Link it from the README (docs list, plus right after Quickstart) and `docs/index.md`
- Add a "Learn more" link to every empty-state panel in the dashboard except Settings (which has no single relevant doc), so each section points at a relevant doc instead of only Launcher Bundles having one

## Why

The dashboard has a lot of distinct features and no single place explained which one to reach for in a given situation. This adds that routing doc and makes it discoverable from the app itself.

## Review notes

Addressed @mxcl's review: corrected the Launcher Bundle row (scripts use Blessing, not a Bundle, Launcher Bundles only wrap unsigned compiled CLIs), tightened the Secret Proxy row's compatibility and security-boundary language to match the narrower canonical claim, and reworded the Temporary Access Grant row to reflect its actual scoping instead of "secret access."

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mxcl:
Good direction—the decision guide and dashboard discoverability are useful. I think the security guidance needs a few corrections before merge:

1. `docs/choosing-a-mechanism.md:24,32-44` says unsigned scripts need Launcher Bundles. Launcher Bundles support one unsigned Mach-O CLI executable; the README explicitly says scripts are unsupported. Scripts use Blessings under a separate Verified Launcher identity.

2. `docs/choosing-a-mechanism.md:22` overstates Secret Proxy compatibility and its security boundary. It requires proxy/CA-compatible HTTP/S software and does not support arbitrary secret files. “A compromised app never sees the raw secret” also conflicts with `docs/positioning.md`’s claim boundary. The narrower canonical claim is that Automic Vault does not place raw Secret values in the launched Target; Secret References and the Proxy Credential remain bearer values, and response redaction is not a general guarantee.

3. `docs/choosing-a-mechanism.md:18` says `av harden` creates an Authorization Gate. Hardening moves a Tool into Hardened State and routes supported operations through its Tool-specific Authorization Gate; hardening state is not runtime authorization evidence.

4. `docs/choosing-a-mechanism.md:26` presents a Temporary Access Grant as generic ten-minute Secret access. It is ten active minutes of narrowly scoped Write Access for one eligible agent task, Tool-specific Gate, Verified Launcher, and runtime posture, with Direct Secret Access, Secret mutation, Elevated Secret Application, Secret Disclosure, and Unknown operations excluded.

5. `MainWindow.swift` maps `.settings` to `nil`. A search with no matching Settings rows renders the shared empty-state panel without the “Learn more” link, despite the PR’s “every empty-state panel” requirement.

The README/index placement and the rest of the SwiftUI link mapping look good.

technicalpickles:
Thanks @mxcl, all five points landed. Fixed:

1. Launcher Bundle row no longer implies script support: reworded to "unsigned compiled CLI" and added that scripts get authority through a Blessing instead.
2. Secret Proxy row now names the actual compatibility requirement (Node 24+ with `--use-env-proxy`, or a client that respects the supplied proxy/CA vars) and swapped "never sees the raw secret" for the narrower claim: no raw Secret value lands in the Target's environment, with the Secret Reference/Proxy Credential still noted as bearer values.
3. Left the hardening/Authorization Gate line as-is, it matches README's own wording ("Hardening adds an Authorization Gate for the Tool"), but added a clause that Hardened State isn't itself a safety guarantee, since that nuance from domain-language.md is worth having either way.
4. Temporary Access Grant row now says "write access to an already-hardened tool," with the actual exclusion list instead of "secret access."
5. Softened the PR description's "every empty-state panel" claim to note the Settings exception, since `.settings` intentionally has no single doc to point at.

Also ran a second, independent pass after that: had a subagent cold-read the doc against nothing but its own text to list every claim that could need checking, then fanned out separate agents to verify each one against README, architecture.md, secret-proxy.md, direct-secret-access.md, signed-cli-launchers.md, and domain-language.md. That caught six more gaps, now fixed: the operation taxonomy and Doctor's checklist were each missing an item, the Temporary Access Grant exclusion list was incomplete and misnamed the Direct Secret Gate, Direct Secret Access was missing its can't-disclose/list/mutate caveat, the two-axes section conflated a Blessing with its separate Launcher Endorsement, and the exits-vs-long-running test didn't account for reentrant Blessed Scripts.

Ready for another look whenever you have time.

mxcl (changes requested):
Thanks for the follow-up. The Secret Proxy and Temporary Access Grant corrections look good, and the explicit Settings exception resolves the empty-state spec mismatch. A few security-language issues remain:

1. `docs/choosing-a-mechanism.md:23` says Direct Access “still can't disclose the raw Secret.” That is false in plain-language outcome terms: Direct Access lets the Launcher choose any Target and arguments, and that Target receives and controls the Secret. For example, it could select a Target that prints its environment. The narrower accurate claim is that a Direct Access Rule does not authorize the distinct Secret Disclosure operation, listing Secret Names, or Secret mutation.

2. `docs/choosing-a-mechanism.md:49-51` now says an unsigned compiled CLI that exits after one reviewed run uses a Blessing. Blessings apply to scripts, not compiled Mach-O executables. This appears to be a regression introduced while fixing the earlier script/Launcher Bundle conflation.

3. `docs/choosing-a-mechanism.md:18` still says `av harden` creates an Authorization Gate and that Hardened State makes its policy apply. The canonical architecture separates these concepts: a Hardener moves a supported Tool into Hardened State; runtime authorization consumes static Gate definitions and performs live checks. Hardened State is neither the gate nor runtime authorization evidence. The README's shorthand defers to `docs/domain-language.md` and `docs/architecture.md`, not vice versa.

4. `docs/choosing-a-mechanism.md:24` still describes Launcher Bundles as the answer for any unsigned CLI. Please constrain this row to a single-file unsigned Mach-O CLI executable; scripts and directory-shaped tools are unsupported.

Two smaller accuracy improvements: qualify `av scan` as covering supported Exposures and Hazards rather than everything on the machine, and describe a Blessing as binding the complete Script Declaration rather than only path and contents.

The README/index placement and dashboard documentation links look good.

technicalpickles:
Accidentally hit re-review without pushing, will need to get it pushed later tonight.

mxcl (approved):
Re-reviewed the latest head after the security-language corrections. Direct Access now requires a Verified Launcher; Temporary Access Grant scope includes the exact Agent Task Context and initial/extendable duration; Gate definitions and durable policy are separated; Launcher Bundle eligibility is constrained to supported unsigned single-file Mach-O CLIs; and the README documentation-list link is restored. No remaining review findings.

PR #306, opened 2026-09-05
Queue human approvals without blocking unrelated secret access (#62)

### LINKED ISSUE #62: Queue manual approval requests without blocking unrelated secret access
## Problem

While a native approval dialog is open, later requests that also require human approval cannot begin. They are held behind the existing modal prompt, even when they concern a different secret or command. Requests that already qualify for automatic approval continue to work.

This creates a denial-of-service condition for manually approved work: one unanswered prompt can prevent all later manually approved secret operations from being presented or decided.

## Expected

A pending request should occupy one approval slot without blocking independent secret requests. Later manual requests should be queued and presented after the current decision, while automatically approved requests continue normally.

## Notes

This does not bypass access control: if nobody is available to decide the visible prompt, later manual work cannot proceed anyway. The goal is to preserve independent access and avoid a global modal bottleneck when approvals are actively handled.

## Investigation

The current manual-approval path dispatches onto the main queue and enters an AppKit modal loop, so later main-queue approval work does not enter while the first dialog is pending. XPC handlers for automatically approved operations load, audit, and reply without waiting on that prompt.

sabrimete:
Resolves #62.

When an interactive Human Approval prompt is displayed, `NSApp.runModal(for: panel)` took over the main run loop exclusively in `NSModalPanelRunLoopMode`. As a result, standard `DispatchQueue.main` and `Task { @MainActor in ... }` tasks scheduled for the default run loop mode were stalled. This caused:
1. **Delayed Auto-Approval Toasts:** Automatic access notifications (`Task { @MainActor in self?.didRecordAccessRequest(record) }`) queued up behind unanswered dialogs and only appeared after the prompt was dismissed.
2. **Head-of-Line Blocking:** Subsequent access requests requiring human approval dispatched their presentation to `DispatchQueue.main` and blocked completely, preventing multiple requests from even queueing properly.

### Changes
- **`HumanApprovalQueue` in `MenubarHelperCore`:** Introduces a thread-safe FIFO queue actor ensuring only one interactive approval prompt occupies the active presentation slot at any given time, while allowing concurrent callers to await or cancel their turn without blocking threads.
- **Decouple from `NSApp.runModal` in `MenubarHelper`:** Refactors `showApprovalAlert` to run non-modally using `withCheckedContinuation`. The main run loop continues spinning in default mode, allowing auto-approval toasts, animations, and IPC handlers to run concurrently.
- **Async Approval Call Sites:** Converts `handleInject`, `handleAccessRequest`, `handleList`, `performApprovedSecretMutation`, and `SecretProxyCoordinator.DestinationApproval` to async/await.
- **Multi-Observer Cancellation:** Enhances `ApprovalCancellation` with UUID-tokened observation so requests waiting in the queue can cancel early before ever presenting a dialog.
- **Prompt Lifecycle & Session Inactivity:** Routes session lock (`userSessionDidResignActive`), screen sleep (`screensDidSleep`), helper reset, and app update (`beginUpdating`) through `ActiveApprovalPrompt.abort()` / `abortActiveApprovalPrompt()`.
- **Interrupted Reply Protocol:** Routes unacquired and aborted queue waiters through `terminalApprovalDecision(.canceled, cancellation: cancellation)` so live clients receive an `interrupted` error reply instead of hanging until timeout.
- **Promotion Race Prevention:** Captures and validates `ActiveApprovalPrompt.abortGeneration` upon slot acquisition to guarantee promoted waiters do not slip past a lock/sleep abort.
- **Keyboard Access:** Sets `panel.initialFirstResponder = panel.contentView`, calls `panel.makeKey()`, and calls `panel.makeFirstResponder(panel.contentView)` upon presentation so Escape (deny) and Return (approve) work immediately without stealing application focus.
- **Tests & Self-Checks:**
  - Expanded unit test suite in `HumanApprovalQueueTests.swift` (8 tests) covering FIFO ordering, cancellation while queued, pre-cancellation, `cancelAllPending`, 30-task concurrent mutual exclusion stress testing, and interleaved cancellation races using bounded deadline-based polling.
  - Adds `abortGeneration`, queue-drain, and `ApprovalPanel` configuration validation in `runApprovalSelfCheck()`.

### Verification
- `swift test --package-path src/menu-helper --filter HumanApprovalQueueTests` (8/8 passed)
- `./scripts/test-menu-helper-self-checks.sh` (all 28 passed)
- `cargo test --tests` (all passed)

### DIFF
--- a/src/menu-helper/Sources/MenubarHelper/main.swift
+++ b/src/menu-helper/Sources/MenubarHelper/main.swift
@@ -195,9 +195,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         isUserSessionActive = false
         temporaryAccessGrants.cancelAll()
         refreshTemporaryAccessGrants()
-        if NSApp.modalWindow is ApprovalPanel {
-            NSApp.abortModal()
-        }
+        abortActiveApprovalPrompt()
     }
 
     @objc private func userSessionDidBecomeActive(_ notification: Notification) {
@@ -209,9 +207,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         areScreensAwake = false
         temporaryAccessGrants.cancelAll()
         refreshTemporaryAccessGrants()
-        if NSApp.modalWindow is ApprovalPanel {
-            NSApp.abortModal()
-        }
+        abortActiveApprovalPrompt()
     }
 
     @objc private func screensDidWake(_ notification: Notification) {
@@ -394,9 +390,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         refreshLiveSecretUses()
         liveSecretUseTimer?.invalidate()
         liveSecretUseTimer = nil
-        if NSApp.modalWindow is ApprovalPanel {
-            NSApp.abortModal()
-        }
+        abortActiveApprovalPrompt()
         automaticApprovalFlashWorkItem?.cancel()
         automaticApprovalFlashWorkItem = nil
         preFlashStatusImage = nil
@@ -519,9 +513,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private func beginUpdating(with alert: NSAlert) -> Bool {
         temporaryAccessGrants.cancelAll()
         refreshTemporaryAccessGrants()
-        if NSApp.modalWindow is ApprovalPanel {
-            NSApp.abortModal()
-        }
+        abortActiveApprovalPrompt()
         let mainWindowWasVisible = mainWindow?.isVisible == true
         mainWindow?.orderOut(nil)
         isUpdating = true
@@ -2613,7 +2605,7 @@ private func performApprovedSecretMutation(
     perform: ((SecretMutation) -> OSStatus)? = nil,
     preflight: (() -> String?)? = nil,
     requestOverride: ApprovalRequest? = nil
-) -> (status: OSStatus?, error: String?) {
+) async -> (status: OSStatus?, error: String?) {
     let request = requestOverride ?? mutation.approvalRequest(callerPath: callerPath)
     if cancellation?.isCanceled == true {
         _ = onAccessRequest(canceledAccessRequestRecord(
@@ -2633,18 +2625,22 @@ private func performApprovedSecretMutation(
         return (nil, "secret mutation denied while user session is inactive")
     }
 
-    let approval = decision?(request) ?? showApprovalAlert(
-        request: request,
-        callerPath: callerPath,
-        pid: pid,
-        signing: signing,
-        scriptApproval: nil,
-        launcher: launcher,
-        launcherFallbackPath: launcherFallbackPath,
-        automaticApprovalExplanation: nil,
-        cancellation: cancellation,
-        compact: mutation.usesCompactApproval
-    )
+    let approval = if let decision {
+        decision(request)
+    } else {
+        await showApprovalAlert(
+            request: request,
+            callerPath: callerPath,
+            pid: pid,
+            signing: signing,
+            scriptApproval: nil,
+            launcher: launcher,
+            launcherFallbackPath: launcherFallbackPath,
+            automaticApprovalExplanation: nil,
+            cancellation: cancellation,
+            compact: mutation.usesCompactApproval
+        )
+    }
     if approval == .interrupted {
         _ = onAccessRequest(interruptedAccessRequestRecord(
             request: request, callerPath: callerPath, launcher: launcher
@@ -2728,38 +2724,51 @@ private func approvalDecision(
 final class ApprovalCancellation: @unchecked Sendable {
     private let lock = NSLock()
     private var canceled = false
-    private var observer: (@MainActor @Sendable () -> Void)?
+    private var observers: [UUID: @MainActor @Sendable () -> Void] = [:]
 
     var isCanceled: Bool {
         lock.withLock { canceled }
     }
 
     func cancel() {
-        let observer: (@MainActor @Sendable () -> Void)? = lock.withLock {
-            guard !canceled else { return nil }
+        let list: [@MainActor @Sendable () -> Void] = lock.withLock {
+            guard !canceled else { return [] }
             canceled = true
-            defer { self.observer = nil }
-            return self.observer
+            let items = Array(self.observers.values)
+            self.observers.removeAll()
+            return items
         }
-        if let observer {
+        if !list.isEmpty {
             RunLoop.main.perform(inModes: [.modalPanel, .default]) {
-                MainActor.assumeIsolated { observer() }
+                MainActor.assumeIsolated {
+                    for item in list {
+                        item()
+                    }
+                }
             }
         }
     }
 
-    func observe(_ observer: @escaping @MainActor @Sendable () -> Void) -> Bool {
+    @discardableResult
+    func observe(id: UUID = UUID(), _ observer: @escaping @MainActor @Sendable () -> Void) -> Bool {
         lock.withLock {
             guard !canceled else { return false }
-            self.observer = observer
+            observers[id] = observer
             return true
         }
     }
 
-    func stopObserving() {
-        lock.withLock { observer = nil }
+    func stopObserving(id: UUID? = nil) {
+        lock.withLock {
+            if let id {
+                observers.removeValue(forKey: id)
+            } else {
+                observers.removeAll()
+            }
+        }
     }
 }
+
 
 private func isApprovalCancellationEvent(_ event: xpc_object_t) -> Bool {
     xpc_equal(event, XPC_ERROR_CONNECTION_INTERRUPTED)
@@ -3625,7 +3634,7 @@ private final class ApprovalServer: @unchecked Sendable {
             )
             return
         }
-        DispatchQueue.main.async {
+        Task { @MainActor in
             if cancellation.isCanceled {
                 _ = self.onAccessRequest(canceledAccessRequestRecord(
                     request: request, callerPath: callerPath, launcher: launcher
@@ -3644,7 +3653,7 @@ private final class ApprovalServer: @unchecked Sendable {
                 self.reply(peer, to: message, ok: false, error: "list denied while user session is inactive")
                 return
             }
-            let decision = showApprovalAlert(
+            let decision = await showApprovalAlert(
                 request: request,
                 callerPath: callerPath,
                 pid: pid,
@@ -4357,7 +4366,7 @@ private final class ApprovalServer: @unchecked Sendable {
                 self.sendEvent(event, to: peer)
             }
         }
-        DispatchQueue.main.async {
+        Task { @MainActor in
             if cancellation.isCanceled {
                 _ = self.onAccessRequest(canceledAccessRequestRecord(
                     request: request, callerPath: callerPath, launcher: promptLauncher
@@ -4483,7 +4492,7 @@ private final class ApprovalServer: @unchecked Sendable {
                 return
             }
 
-            let decision = showApprovalAlert(
+            let decision = await showApprovalAlert(
                 request: request,
                 callerPath: callerPath,
                 pid: pid,
@@ -4901,7 +4910,7 @@ private final class ApprovalServer: @unchecked Sendable {
                     self.sendEvent(humanApprovalRequiredEvent, to: peer)
                 }
             }
-            DispatchQueue.main.async {
+            Task { @MainActor in
                 if cancellation.isCanceled {
                     _ = self.onAccessRequest(canceledAccessRequestRecord(
                         request: request, callerPath: callerPath, launcher: launcher
@@ -4920,7 +4929,7 @@ private final class ApprovalServer: @unchecked Sendable {
                     self.reply(peer, to: message, ok: false, error: "human approval unavailable")
                     return
                 }
-                let decision = showApprovalAlert(
+                let decision = await showApprovalAlert(
                     request: request,
                     callerPath: callerPath,
                     pid: pid,
@@ -5110,12 +5119,12 @@ private final class ApprovalServer: @unchecked Sendable {
         let warning = targetProtection?.allowsSecretGateAccess == true ? nil :
             "The target does not meet Automic Vault’s Hardened Runtime requirements. Code injected into it may steal this Proxy Session’s references and credential, then reuse destinations you allow for the session."
 
-        DispatchQueue.main.async {
+        Task { @MainActor in
             guard !cancellation.isCanceled, self.canRequestHumanApproval() else {
                 self.reply(peer, to: message, ok: false, error: "Proxy Session approval unavailable")
                 return
             }
-            let decision = showApprovalAlert(
+            let decision = await showApprovalAlert(
                 request: request,
                 callerPath: callerPath,
                 pid: pid,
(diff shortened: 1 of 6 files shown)


mxcl:
Thanks for working on this. We’ve confirmed the scheduling issue, but we’re not yet certain how best to proceed. We’ll get back to you once we’ve considered the approach further.

src/menu-helper/Sources/MenubarHelper/main.swift
```diff
@@ -11324,99 +11415,95 @@ private func showApprovalAlert(
...
+        if !observed {
+            state.resolve(.canceled)
+            return
+        }
+
+        fitApprovalPanel(panel, maximumHeight: maximumHeight, animate: false)
+        panel.center()
+        panel.orderFrontRegardless()
+        panel.makeKey()
+        panel.makeFirstResponder(panel.contentView)
```
mxcl: **[P1] Preserve deliberate focus before enabling keyboard approval.** Together with changing `ApprovalPanel.canBecomeKey` to always return true, these calls remove the existing click-before-focus protection and make every newly presented prompt keyboard-active. `ApprovalPromptApprovalMenu` already binds Return through `.keyboardShortcut(.defaultAction)`, and ordinary Mac approval directly calls `decide(.approved)`. A Return intended for another window, or a repeated Return as the next queued prompt appears, can therefore approve a request the user has not reviewed. The isolated AppKit probe confirmed that the base panel is not key before a click, whereas this version is. Please retain deliberate focus acquisition when removing the modal loop and cover the transition between queued prompts. Making notification delivery responsive does not require this change in approval input behavior.

src/menu-helper/Sources/MenubarHelper/main.swift
```diff
@@ -11324,99 +11415,95 @@ private func showApprovalAlert(
...
+                content: content,
+                maximumHeight: maximumHeight,
+                allowsPersistentApproval: allowsPersistentApproval,
+                temporaryGrantCandidate: temporaryGrantCandidate,
+                persistentApprovalLabel: persistentApprovalLabel,
+                usesIPhoneApproval: usesIPhoneApproval,
+                usesTouchIDApproval: usesTouchIDApproval,
+                compact: compact,
+                decide: { userDecision in
+                    state.resolve(userDecision)
```
mxcl: **[P1] Invalidate pending prompts when the required approval surface changes.** These flags capture the settings at presentation, but the new nonmodal lifetime allows Settings to be used while this prompt remains open. If an ordinary Mac prompt is pending and the user enables iPhone Approval or Touch ID Approval, the existing view retains its pointer/keyboard allow action. Neither enable path cancels or refreshes that prompt, and `state.resolve` does not validate the currently permitted approval surface. This conflicts with the settings promise that Mac allow actions disappear or fresh Touch ID becomes required. Please cancel/recreate affected prompts on approval-mode transitions, or enforce the current permitted surface before accepting an allow result. This is based on tracing `MainWindow.swift` and `PhoneApprovalCoordinator.swift`; live enrollment has not been exercised.

src/menu-helper/Sources/MenubarHelper/main.swift
```diff
@@ -4483,7 +4492,7 @@ private final class ApprovalServer: @unchecked Sendable {
                 return
```
mxcl, line 4689: **[P2] Reevaluate authorization after a queued request gets its turn.** The recent-decision cache and Temporary Access Grants are checked before this call waits for the presentation slot. With the main queue now free, multiple requests arriving during the first prompt can all miss those checks and enqueue. If the first decision then establishes a reusable approval/denial or a Temporary Access Grant, the queued requests proceed directly to another prompt without reevaluation. Previously their main-queue callbacks ran after the first decision and could use the new authorization state. This produces repeated prompts for identical requests and unnecessary prompts for writes newly covered by a grant. Please reevaluate the applicable decision reuse and grants after acquiring the slot, before presenting, and add a test through the approval callsite rather than only the queue primitive.

src/menu-helper/Sources/MenubarHelperCore/HumanApprovalQueue.swift
```diff
@@ -0,0 +1,129 @@
...
+                resumed = true
+                continuation.resume(returning: granted)
+            }
+
+            let wasAlreadyCanceled: Bool = lock.withLock {
+                if isCanceled?() == true {
+                    return true
+                }
+                waiters[id] = resumeWith
+                waiterOrder.append(id)
```
mxcl: **[P1] Make the active-slot check and waiter registration atomic.** `acquire` checks `isSlotActive` under the earlier lock, releases that lock, and only registers the continuation here. An active holder can call `release()` between those steps, find no waiters, and clear `isSlotActive`. This code then enqueues a waiter without rechecking the slot, leaving it suspended with nobody responsible for waking it. I reproduced this with the unmodified Swift 6 queue source in four out of four stress runs: repeated groups of 32 tasks immediately acquiring and releasing the slot hung with `hasActiveSlot == false` and `pendingCount == 1` or `2`. Please combine the decision to acquire or enqueue under one lock and add coverage for release racing with waiter registration. The eight existing tests pass but do not catch this interleaving.

mxcl (changes requested):
Thanks for working on this. Decoupling approval presentation from the main dispatch queue is a reasonable direction, but these changes need further work before we can merge them. I found a reproducible queue hang, two approval-surface safety concerns, and a regression in reevaluating queued requests. Details are inline.

Validation: the eight added queue tests passed in an isolated Swift package using the unmodified queue and test sources. A separate stress probe using the unmodified queue hung in four out of four runs. An isolated AppKit probe confirmed the change in keyboard focus. The approval-mode transition and queued-request reevaluation findings are based on source tracing; I have not exercised live iPhone or Touch ID enrollment.

src/menu-helper/Sources/MenubarHelperCore/HumanApprovalQueue.swift
```diff
@@ -0,0 +1,129 @@
...
+        lock.withLock {
+            isSlotActive = false
+            for resume in waiters.values {
+                resume(false)
+            }
+            waiters.removeAll()
+            waiterOrder.removeAll()
+        }
+    }
+}
```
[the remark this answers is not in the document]
sabrimete: Fixed in bb069e7a. `resetForTesting()` now mirrors `cancelAllPending()`: it snapshots the registered continuations, clears state under `lock`, and resumes the continuations outside the lock. Added unit test `resetForTestingDoesNotDeadlockWhenWaitersReenter` to verify re-entrant calls do not deadlock.

src/menu-helper/Sources/MenubarHelper/main.swift
```diff
@@ -11324,99 +11415,95 @@ private func showApprovalAlert(
...
+                state.remoteRequestID = phoneRequest.id
+                try PhoneApprovalCoordinator.shared.submit(phoneRequest) { result in
+                    let mapped: ApprovalDecision = switch result {
+                    case .approved: .approved
+                    case .denied: .denied
+                    case .temporaryWriteAccess: .temporaryWriteAccess
+                    case .canceled: .canceled
+                    }
+                    state.resolve(mapped)
                 }
```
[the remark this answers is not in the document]
sabrimete: Fixed in bb069e7a. Wrapped the `PhoneApprovalCoordinator.shared.submit` completion and `alertRequest.observeCancellation` callbacks in explicit `Task { @MainActor in state.resolve(..., source:) }` blocks to ensure compiler-enforced `@MainActor` isolation.

src/menu-helper/Sources/MenubarHelperCore/HumanApprovalQueue.swift
sabrimete: Fixed in bb069e7a. `acquire()` now checks slot availability, pre-cancellation, and registers waiters atomically under a single lock inside the continuation. If the slot is free, it marks `isSlotActive = true` and returns immediately; if occupied, it registers the waiter before unlocking. Added test `releaseRacingWithWaiterRegistration` executing 5 consecutive waves of 32 concurrent tasks acquiring and releasing the slot without drops or hangs.

src/menu-helper/Sources/MenubarHelper/main.swift
sabrimete: Restored and verified in bb069e7a and 72b4a00f.
- Reverted `ApprovalPanel.canBecomeKey` to guard on an explicit mouse click (`canBecomeKey { allowsKey }` with `allowsKey = false` until `.leftMouseDown`).
- Removed `makeKey()` and `makeFirstResponder(nil)` from presentation setup.
- Added assertions in `--self-check-approval-callsite` verifying that both initially presented panels and subsequent queued panels presented after dismiss remain non-key (`!canBecomeKey`, `!isKeyWindow`, `initialFirstResponder == nil`), confirming focus is not inherited across queued prompts without deliberate user interaction.

src/menu-helper/Sources/MenubarHelper/main.swift
sabrimete: Addressed in bb069e7a and refined in 72b4a00f:
- Active prompts are aborted on surface configuration transitions (`TouchIDApproval.enable/disable`, `PhoneApprovalCoordinator.enable/disable/recover`).
- In `ApprovalPromptState.resolve(_:source:)`, decision sources are validated against current elevation requirements. If an elevated factor is configured, standard Mac approval clicks are coerced to `.interrupted` (failing closed without poisoning the transient denial cache). When both iPhone Approval and Touch ID are enabled, either strong factor (`.phone` or `.touchID`) is accepted.
- Verified hermetically in `--self-check-approval-callsite` across all four surface permutations (`standardMacOnly`, `touchIDOnly`, `phoneOnly`, `bothElevated`).

src/menu-helper/Sources/MenubarHelper/main.swift
sabrimete: Implemented in bb069e7a and verified end-to-end in 72b4a00f:
- `showApprovalAlert` now accepts an async `reevaluate: () -> Bool` closure invoked immediately upon acquiring the presentation slot. If covered by a newly granted access token or transient approval cache entry, it releases the slot and returns `.reevaluated`, bypassing prompt presentation entirely.
- Added `--self-check-approval-callsite` (wired into `scripts/test-menu-helper-self-checks.sh`), which runs request 1 holding the slot, enqueues request 2, establishes an approved decision in `AuthorizationDecisionReuseCache` using production key derivation (`testReq.decisionReuseRequest(clientIdentity: selfIdentity, ...)`), releases the slot, and verifies request 2 resolves with `.reevaluated` without presenting a UI panel.

sabrimete:
Addressed all review feedback from @mxcl and @copilot across 2 focused commits (`bb069e7a` and `72b4a00f`):

1. **Atomic Queue Acquisition & Re-entrant Deadlock Prevention**:
   - Combined `isSlotActive` inspection and waiter registration into a single lock acquisition in `HumanApprovalQueue.acquire`, eliminating the race where `release()` could miss pending waiters.
   - Snapshot continuations under lock in `resetForTesting()` before resuming them outside lock.
   - Added unit tests `releaseRacingWithWaiterRegistration` (5 waves × 32 concurrent tasks) and `resetForTestingDoesNotDeadlockWhenWaitersReenter`.

2. **Deliberate Focus Protection**:
   - Restored click-before-focus protection on `ApprovalPanel` (`canBecomeKey` guarded by explicit `.leftMouseDown`).
   - Removed `makeKey()` / `makeFirstResponder()` calls on initial presentation.

3. **Approval Surface Transitions & Enforcement**:
   - Active prompts are invalidated when Touch ID or Phone Approval modes are toggled.
   - `ApprovalPromptState.resolve` validates decision sources against configured security posture. When elevated factors are active, standard Mac button clicks are fail-closed to `.interrupted` (preventing cached denial poisoning). When both Touch ID and Phone Approval are active, either strong factor is accepted.

4. **Queued Request Reevaluation**:
   - `showApprovalAlert` reevaluates Temporary Access Grants and recent decisions upon slot acquisition, returning `.reevaluated` and skipping prompt presentation if already authorized.

5. **End-to-End Callsite Self-Check**:
   - Added `--self-check-approval-callsite` to `test-menu-helper-self-checks.sh` testing slot handoff, production key-derived decision cache reevaluation, queued-transition focus isolation, and all 4 surface elevation permutations.

src/menu-helper/Sources/MenubarHelper/main.swift
```diff
@@ -14483,6 +14660,248 @@ private func runApprovalSelfCheck() -> Int32 {
...
+        group.addTask {
+            await queuedAlertTask.value
+        }
+        group.addTask {
+            try? await Task.sleep(for: .seconds(5))
+            return nil
+        }
+        let first = await group.next() ?? nil
+        group.cancelAll()
+        return first
```
mxcl: **[P2] Make the self-check timeout terminate a stuck approval.** If the five-second timer wins, `group.cancelAll()` does not interrupt the other child awaiting `queuedAlertTask.value`: that is an unstructured task, and its value wait does not automatically respond to cancellation. Leaving the task-group scope still waits for that child, so a stuck approval makes this self-check hang instead of returning the intended failure code 44. I reproduced the same control flow with a standalone probe; an independent watchdog fired after the group timeout had elapsed. Please explicitly cancel/resolve the pending approval through its cancellation path before leaving the group, or enforce the timeout outside the process. Add a negative-path check where the approval never completes. This is a test robustness issue; the earlier four findings were addressed in the version reviewed at `72b4a00f`.

sabrimete: Addressed in 07bb7b2e:

1. **Explicit Cancellation on Timeout:** Introduced `@MainActor awaitWithTimeout(duration:cancellation:task:)`, requiring `ApprovalCancellation` so callers must explicitly provide the cancellation path. When the timeout child completes first (`first == nil`), it calls `cancellation.cancel()` and `task.cancel()` before `group.cancelAll()` and draining the remaining child, ensuring the child awaiting `task.value` is unblocked before leaving the task-group scope.
2. **Cancellation Dispatch Under `dispatchMain()`:** Fixed `ApprovalCancellation.cancel()` to dispatch off-main cancellations to `DispatchQueue.main.async` rather than `RunLoop.main.perform(inModes:)`. Under the self-check's `dispatchMain()` entry there is no active `NSRunLoop`, which prevented the previous perform block from firing and caused the unblock to stall. `DispatchQueue.main.async` executes promptly under both `dispatchMain()` and AppKit modal run loops, while on-main cancellations continue firing synchronously under `MainActor.assumeIsolated`.
3. **Never-Completes Negative Path:** Added Section 5 (codes 50–53) to `runApprovalCallsiteSelfCheck`. It holds the active slot, enqueues an approval with `reevaluate: { false }`, and verifies that `awaitWithTimeout(duration: .milliseconds(500), cancellation: ...)` cleanly aborts, returning `nil` and leaving the queue fully drained after slot release.
4. **Non-Visible Panel Guard:** Guarded `makeKey()` in `ApprovalPanel.sendEvent` with `if isVisible` and closed test panels in Section 1 to prevent headless test alerts from scheduling window-server display passes.

All 29 self-checks and test suites pass cleanly.

mxcl (approved):
Reviewed 07bb7b2e. The four earlier findings and the self-check timeout issue are resolved, with no remaining actionable findings. The local build, all 29 self-checks, and all 10 queue tests passed; Swift and Rust CI are now green. Live phone and biometric transitions were not exercised.

PR #307, opened 2026-09-05
Detect plaintext Gemini CLI and Antigravity Cockpit credentials

sabrimete:
### Summary

Adds dedicated detect-only isotopes for `gemini-cli` (Gemini CLI) and `antigravity-cockpit` (Antigravity Cockpit):

1. **`gemini-cli`**: Checks `~/.gemini/oauth_creds.json` (mode 0600) for cached OAuth credentials (`access_token`, `refresh_token`, `id_token`).
2. **`antigravity-cockpit`**: Checks `~/.antigravity_cockpit/credentials.json` (mode 0644, world-readable) for legacy stored account credentials (`accessToken`, `refreshToken`, `expiresAt`).

Both credential files store live refresh tokens that persist beyond access-token expiry. Neither tool currently has an Automic Vault hardener, and both detectors provide clear, actionable remediation guidance.

### Changes
- Add `src/isotopes/detectors/gemini_cli/` detector and documentation.
- Add `src/isotopes/detectors/antigravity_cockpit/` detector and updated documentation referencing Cockpit v2.1.29 migration to VS Code `SecretStorage`.
- Register `gemini-cli` and `antigravity-cockpit` in `src/isotopes/detectors/mod.rs` in alphabetical order.
- Distinguish file absence from traversal/metadata inspection errors using `Path::try_exists()`.
- Add unit tests covering parseable, partial, invalid, unreadable, and unsearchable parent directories.
- Add integration tests in `tests/scan_cli.rs` covering targeted scans and unsearchable directory inspection failures.

### Verification
- `cargo fmt --check` (passes)
- `cargo test isotopes::detectors` (519 passed, 0 failed)
- `cargo test --test scan_cli` (8 passed, 0 failed)
- Verified with live credential files and simulated permission denial on parent directories.


src/isotopes/detectors/gemini_cli/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+}
+
+pub(crate) fn oauth_path() -> Result<PathBuf, String> {
+    oauth_path_in(std::env::var_os("HOME").as_deref())
+}
+
+fn oauth_path_in(home: Option<&OsStr>) -> Result<PathBuf, String> {
+    home.filter(|value| !value.is_empty())
+        .map(|home| PathBuf::from(home).join(".gemini/oauth_creds.json"))
```
mxcl, line 48: [P2] Establish credential ownership before reporting this as agy

This is Gemini CLI's OAuth credential path: its Storage.getOAuthCredsPath() resolves to ~/.gemini/oauth_creds.json (https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/config/storage.ts). Because this detector unconditionally labels matches as agy, a Gemini-only installation produces an Antigravity finding. Google's Antigravity documentation also describes ordinary local sign-in using the OS keyring, contradicting detector.md's claim that developer OAuth cannot use Keychain: https://antigravity.google/docs/cli/install/.

Please substantiate the supported Antigravity version and its actual credential file, and align the detector and documentation with that evidence. If an Antigravity-specific exposure cannot be established, remove this detector from the PR. A Gemini detector can cover this file separately under the correct name.

src/isotopes/detectors/agy/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    install_insecurity_reasons().map(|reasons| !reasons.is_empty())
+}
+
+pub fn install_insecurity_reasons() -> Result<Vec<String>, String> {
+    let oauth = oauth_path()?;
+    Ok(oauth_insecurity_reasons_for(&oauth))
+}
+
+fn oauth_insecurity_reasons_for(path: &Path) -> Vec<String> {
+    if !path.exists() {
```
mxcl: [P2] Distinguish inspection errors from missing files in both detectors

Path::exists() returns false when metadata inspection fails, so this guard bypasses the unreadable-file finding when a parent directory cannot be searched. Reproduced with the PR binary and synthetic credentials for both agy and antigravity-cockpit: each reports one finding while readable; after chmod 000 on its credential directory, stat raises PermissionError but av scan --json --detector <name> succeeds with {"findings":[]} although the credential file still exists.

Distinguish absence from inspection failure (for example, handle try_exists() errors), and add a focused regression check for an unsearchable parent. Apply the same correction to cockpit_insecurity_reasons_for. The existing directory-at-file-path test does not exercise this failure.

src/isotopes/detectors/antigravity_cockpit/detector.md
```diff
@@ -0,0 +1,38 @@
...
+`~/.antigravity_cockpit/credentials.json` with permissions mode `0644`
+(world-readable). The file holds live refresh tokens that survive access token
+expiry, creating a severe Exposure that allows any local user or unauthorized
+process on the system to obtain fresh access tokens.
+
+## Why This is not Yet Hardened
+
+Antigravity Cockpit stores mutable account credential state on disk without a
+native keychain-backed credential store or pluggable credential helper. A safe fix
+requires upstream support for system keychain custody or a dedicated source isotope.
```
mxcl: [P2] Replace obsolete Cockpit remediation guidance

Cockpit already stopped reading and writing this shared plaintext credential file in v2.1.29 (March 3, 2026): https://github.com/jlcodes99/vscode-antigravity-cockpit/blob/58d51e3074def47513cb1de6338f76e95b3a638e/CHANGELOG.md#2129---2026-03-03. Its credential storage uses context.secrets and SecretStorage.store: https://github.com/jlcodes99/vscode-antigravity-cockpit/blob/58d51e3074def47513cb1de6338f76e95b3a638e/src/auto_trigger/credential_storage.ts.

Detecting leftover plaintext tokens remains useful, but this paragraph is also surfaced as the finding's solution, so users are incorrectly told to await upstream storage support. Describe the legacy shared cache and the available upgrade/remediation path, including addressing exposed residue, and distinguish that from any missing Automic Vault Hardener. Update the repeated unsupported claims later in this document too.

mxcl (changes requested):
Requesting changes for the three findings below: credential attribution, inspection failures being treated as clean scans, and obsolete Cockpit remediation guidance.

For agy, please establish the exact upstream version, credential path, and storage behavior before retaining this detector. If we cannot substantiate an Antigravity-specific exposure, remove the agy detector, registration, documentation, and associated tests from this PR. The Gemini credential exposure can be covered separately under the correct Tool name; it should not be presented as Antigravity.

Validation at 5f5f9aa7: cargo fmt --check, all 517 detector tests, and all 7 scan CLI tests pass. The directory-permission failure described below was reproduced separately with synthetic credentials for both new detectors.

mxcl:
Clarification on the agy finding: reattributing this detector to **Gemini CLI in this PR** is a sensible fix. The plaintext credential exposure is real; the incorrect part is presenting it as Antigravity. My review was too restrictive in suggesting that Gemini coverage must be a separate change.

Please update the detector/module name, registration, finding text, documentation, and tests to match Gemini CLI, and independently verify the hardening explanation against Gemini CLI's actual credential-storage options. Simply replacing the product name throughout could preserve unsupported claims.

An installation check is unnecessary: leftover credential files remain an exposure even after the tool is uninstalled. Removing the detector is the fallback only if its credential format or ownership cannot be substantiated.

src/isotopes/detectors/gemini_cli/mod.rs
sabrimete: Addressed. The detector has been reattributed to Gemini CLI (`gemini_cli` module, `gemini-cli` detector name) and registered alphabetically in `src/isotopes/detectors/mod.rs`.

The documentation and finding solution have been updated to reflect Gemini CLI's actual credential behavior: default plaintext JSON caching in `~/.gemini/oauth_creds.json` (mode `0600`), the presence of live refresh tokens, its optional encrypted file backend via `GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true`, and the lack of a default native keychain store or pluggable credential helper.

src/isotopes/detectors/agy/mod.rs
sabrimete: Addressed. Both `gemini_cli` and `antigravity_cockpit` now use `path.try_exists()` to distinguish absence from metadata inspection failure:
- `Ok(false)` returns empty findings for confirmed absence.
- `Ok(true)` reads the file and checks for credentials or reports unreadable/unparseable findings.
- `Err(_)` treats inspection errors (such as `PermissionDenied` when a parent directory is unsearchable) as unreadable credential findings rather than silently returning a clean scan.

Added focused unit tests in both detector modules (`reports_when_parent_directory_is_unsearchable`) and an integration test in `tests/scan_cli.rs` (`av_scan_json_reports_unsearchable_credential_parent_directory`).

src/isotopes/detectors/antigravity_cockpit/detector.md
sabrimete: Addressed. Updated `antigravity_cockpit/detector.md` and the finding's solution to describe `~/.antigravity_cockpit/credentials.json` as legacy unencrypted residue from versions prior to v2.1.29.

The mitigation guidance now points users to upgrade to Cockpit v2.1.29+ (which uses VS Code `SecretStorage` backed by the OS Keychain), remove the unencrypted residue file, and consider revoking exposed tokens. Also clarified why Automic Vault does not need to provide a custom hardener given upstream's native secret storage adoption.

sabrimete:
Thanks for the detailed review @mxcl. All three findings have been addressed in the updated commit:

1. **Reattributed to Gemini CLI**:
   - Renamed the detector module to `gemini_cli` and registered it as `gemini-cli` in alphabetical order in `src/isotopes/detectors/mod.rs`.
   - Updated the documentation and finding solution to accurately describe Gemini CLI's default credential path (`~/.gemini/oauth_creds.json`, mode `0600`), the presence of live refresh tokens, its optional encrypted file backend via `GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true`, and the lack of a default native keychain integration or external credential helper interface.

2. **Handled Inspection Errors (`Path::try_exists`)**:
   - Both `gemini_cli` and `antigravity_cockpit` now use `path.try_exists()`. Confirmed absence (`Ok(false)`) returns a clean scan, while traversal and metadata errors (`Err(_)`, such as `PermissionDenied` when a parent directory has mode `000`) report an unreadable credential finding rather than being bypassed.
   - Added focused unit tests in both modules (`reports_when_parent_directory_is_unsearchable`) and an end-to-end integration test in `tests/scan_cli.rs` (`av_scan_json_reports_unsearchable_credential_parent_directory`).

3. **Updated Antigravity Cockpit Remediation Guidance**:
   - Updated `antigravity_cockpit/detector.md` and the finding solution to describe `~/.antigravity_cockpit/credentials.json` as unencrypted legacy residue from pre-2.1.29 installations.
   - Clarified that Cockpit v2.1.29+ moved to VS Code `SecretStorage` backed by the OS Keychain, and updated remediation instructions to direct users to upgrade to v2.1.29+, remove the residue file, and consider revoking exposed tokens.
   - Clarified why Automic Vault does not need to provide a custom hardener given upstream's native secret storage adoption.

All formatting checks, 519 detector unit tests, and 8 scan CLI integration tests pass cleanly.

src/isotopes/detectors/gemini_cli/detector.md
```diff
@@ -0,0 +1,34 @@
...
+permissions mode `0600`. The file contains plaintext `access_token`,
+`refresh_token`, and `id_token` values. Live refresh tokens survive access-token
+expiry, allowing unauthorized processes running under the same user account to
+mint new access tokens and call Gemini services without authentication prompts.
+
+## Mitigation
+
+Delete `~/.gemini/oauth_creds.json` and authenticate via short-lived environment
+credentials, or configure Gemini CLI's encrypted file storage support with
+`GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true`.
```
mxcl: [P2] Qualify the encrypted-file mitigation against the same-user threat

The original attribution, inspection-error handling, and Cockpit remediation findings are addressed; I have resolved those threads. The updated code passes cargo fmt --check, all 519 detector tests, and all 8 scan CLI tests.

This new recommendation needs a security qualification. In Gemini CLI v0.58.0, the encrypted-file fallback derives its AES key entirely from the fixed string `gemini-cli-oauth`, hostname, and username. Another process running as the user can reconstruct the key and recover the credentials; this does not mitigate the same-user credential access described immediately above. See [FileKeychain.deriveEncryptionKey](https://github.com/google-gemini/gemini-cli/blob/v0.58.0/packages/core/src/services/fileKeychain.ts#L22-L25).

The selected storage path prefers native Keychain, but [falls back to FileKeychain when native storage is unavailable](https://github.com/google-gemini/gemini-cli/blob/v0.58.0/packages/core/src/services/keychainService.ts#L90-L107). Setting the flag alone therefore does not establish native Keychain custody. The migration can also remove the original plaintext file, making this detector quiet while the fallback still permits same-user recovery.

Please distinguish native Keychain storage from the encrypted-file fallback and explicitly document the fallback's limitation. Do not present encrypted-file storage alone as resolving this Exposure. This is a correction to the user-facing mitigation, not a request to add a new Hardener.

sabrimete: Addressed in 7467e732.

Updated `gemini_cli/detector.md` to:
1. Advise deleting `~/.gemini/oauth_creds.json` and authenticating via short-lived environment credentials (`GOOGLE_APPLICATION_CREDENTIALS` or `GOOGLE_CLOUD_ACCESS_TOKEN`) as the primary mitigation.
2. Explicitly distinguish native OS Keychain storage from the `FileKeychain` fallback (`~/.gemini/gemini-credentials.json`).
3. Document that `FileKeychain` derives its AES key from static, unprivileged metadata (application name, hostname, and username), allowing any process running under the same user account to reconstruct the key and recover the credentials.
4. Explicitly warn users not to rely on encrypted-file storage alone to resolve same-user credential exposure.

src/isotopes/detectors/gemini_cli/detector.md
```diff
@@ -0,0 +1,40 @@
...
+Gemini CLI caches OAuth credentials in `~/.gemini/oauth_creds.json` with
+permissions mode `0600`. The file contains plaintext `access_token`,
+`refresh_token`, and `id_token` values. Live refresh tokens survive access-token
+expiry, allowing unauthorized processes running under the same user account to
+mint new access tokens and call Gemini services without authentication prompts.
+
+## Mitigation
+
+Delete `~/.gemini/oauth_creds.json` and authenticate via short-lived environment
+credentials such as `GOOGLE_APPLICATION_CREDENTIALS` or `GOOGLE_CLOUD_ACCESS_TOKEN`.
```
mxcl: [P2] Do not describe GOOGLE_APPLICATION_CREDENTIALS as inherently short-lived

GOOGLE_APPLICATION_CREDENTIALS is a path to a JSON credential file, not an access token or a guarantee of credential lifetime. It explicitly supports service-account private keys. Following this mitigation can therefore replace the exposed OAuth file with another plaintext, long-lived credential, preserving the same-user Exposure. This paragraph is also surfaced as the finding's solution.

Please remove this example or explicitly constrain it to an appropriate keyless federation configuration without a locally stored long-lived credential. Google's ADC documentation explains the supported file types and warns about service-account keys: https://docs.cloud.google.com/docs/authentication/application-default-credentials#GAC.

src/isotopes/detectors/gemini_cli/detector.md
```diff
@@ -0,0 +1,40 @@
...
+
+While Gemini CLI supports token storage migration via the OS Keychain, it falls
+back to an encrypted file (`~/.gemini/gemini-credentials.json`) when native
+keychain storage is unavailable or when `GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true`
+is set. This fallback derives its AES key entirely from static metadata (the
```
mxcl: [P2] Name the flag that actually forces file fallback

In Gemini CLI v0.58.0, GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true selects OAuthCredentialStorage, whose backend still prefers native Keychain. The separate GEMINI_FORCE_FILE_STORAGE=true flag forces FileKeychain. The current wording therefore identifies the wrong setting as bypassing native custody.

Please replace the forced-fallback condition with GEMINI_FORCE_FILE_STORAGE=true, or explain the two independent selections explicitly. Native storage being unavailable is already a correct fallback condition.

Sources: [OAuth storage selection](https://github.com/google-gemini/gemini-cli/blob/v0.58.0/packages/core/src/code_assist/oauth2.ts#L104-L137), [Keychain versus file selection](https://github.com/google-gemini/gemini-cli/blob/v0.58.0/packages/core/src/services/keychainService.ts#L90-L107).

mxcl (changes requested):
The encrypted-file warning now correctly explains the same-user threat. Two corrections remain in the newly added Gemini authentication guidance, detailed inline below. These are documentation fixes; no additional Detector or Hardener is requested.

Reviewed the documentation-only update at 7467e732. All 16 detector metadata/documentation tests pass. The detector code is unchanged from the previously validated revision (519 detector tests and 8 scan CLI tests passed).

src/isotopes/detectors/gemini_cli/detector.md
sabrimete: Addressed in dbe728e9. Removed `GOOGLE_APPLICATION_CREDENTIALS` from the mitigation guidance and finding solution, advising authentication via short-lived access tokens (`GOOGLE_CLOUD_ACCESS_TOKEN`) to avoid substituting the OAuth credential with another persistent disk credential.

src/isotopes/detectors/gemini_cli/detector.md
sabrimete: Addressed in dbe728e9. Clarified the two settings: `GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true` activates encrypted storage (which prefers the native Keychain), while `GEMINI_FORCE_FILE_STORAGE=true` or unavailable native storage forces the `FileKeychain` fallback.

sabrimete:
Updated in `dbe728e9` with the following documentation corrections:

1. **Short-Lived Authentication**: Removed `GOOGLE_APPLICATION_CREDENTIALS` from `gemini_cli/detector.md` and the extracted finding solution, advising authentication via short-lived access tokens (`GOOGLE_CLOUD_ACCESS_TOKEN`) rather than pointing to persistent key files.
2. **Storage Selection Flags**: Clarified that `GEMINI_FORCE_ENCRYPTED_FILE_STORAGE=true` enables encrypted storage (which prefers the native Keychain), while `GEMINI_FORCE_FILE_STORAGE=true` or native storage unavailability forces the `FileKeychain` fallback.
3. **Domain Language Alignment**: Capitalized canonical domain terms (`Exposure` and `Secret Gate`) per `docs/domain-language.md`.

All 519 detector unit tests, 16 metadata/documentation tests, and 8 scan CLI tests pass cleanly, and the release build (`./scripts/build.sh --wmo`) verified.

mxcl (approved):

Issue #76, opened 2026-07-25
Detector for `codesign` having always approve


mxcl:
If you have a codesigning identity installed and have always allowed it from the keychain it means malware or agents could codesign as you and replace an existing secret gate approved launcher.

This only applies when secret gates can be added for self-signed tools like in #73.

Also it applies a lot to me since my Apple identity is on my Mac.

mxcl:
A hardener would work for this situation, probably since codesign will accept env vars for the keys.

mxcl:
In fact it may be enough to require notarization. Though this is tedious ofc. We can add knobs so the user can determine the amount of tedium they will tolerate for true security.

PR #327, opened 2026-09-09
Draft: detect permissive code-signing key ACLs (no-prompt blocker)

mxcl:
Code-signing identities can delegate signing to `/usr/bin/codesign` or all applications through their Keychain ACL. This draft reports that configuration as a **medium-severity Hazard**, with a separate incomplete-inspection finding for unreadable or unsupported metadata. It does not provide a Hardener or claim confirmed unattended signing.

**DO NOT MERGE: the no-prompt requirement remains unresolved.**

A full disposable-identity test discovered that `SecKeychainItemCopyAccess` can prompt to unlock a locked Keychain despite `SecKeychainSetUserInteractionAllowed(false)` and a no-UI identity query. A process sample located the request in `SecKeychainItemCopyAccess → CSSM_GetKeyOwner`; it occurred during locked-key inspection, not setup or cleanup. The prompt was canceled, and the fixture was deleted. No real signing key was used or changed.

The draft now checks Keychain status before reading the ACL and reports known-locked keys as incomplete. The revised fixture completes without a prompt. **A Keychain can still lock between that check and the legacy ACL call.** The registration and display-name change remain in this draft so the proposed integration can be reviewed, but they must not ship until that race has a supported, prompt-free solution. The public legacy API's owner-ACL RPC has no UI-control argument; adding another preflight cannot prove the required invariant.

The implementation reads public certificates, identity/key references, and signing ACL metadata. It distinguishes SIGN (115) from DECRYPT (24), ANY, nil/all-app versus empty/no-trusted-app lists, and the exact codesign trusted-app entry. Certificate selection requires explicit Code Signing EKU, including self-signed certificates; it does not evaluate trust, dates, revocation, or network state. Partition constraints and effective authentication remain unverified. No signing, private-key export, credential-tool invocation, or network request occurs in the Detector.

Ordinary Launcher identity still depends on the matching designated requirement and runtime eligibility; a usable matching Developer ID key can undermine that trust. Generated Launcher Bundles additionally check exact enrolled code identifiers, payload digests, and generation data, so merely re-signing changed code does not satisfy their enrollment.

Validation:
- `cargo test --lib codesign`: 4 passed. Native in-memory ACL fixtures, public certificate EKU fixtures, medium/unknown mapping, and a full disposable identity scoped exclusively to its temporary Keychain. Unlocked result: one codesign grant. Known-locked result: exactly one unknown. The fixture restores the interaction setting and explicitly unlocks/deletes only its own Keychain using its fixture password.
- `cargo test --lib isotopes::detectors::tests`: 16 passed.
- `cargo fmt --check` and `git diff --check`: passed.
- No real-key scan or intentionally racy lock test was run. Swift display-name mapping was not separately built locally.

Before release, integrate this non-file state Detector with the periodic refresh opt-in from #326 / #231 when that change is available. This draft does not import that unrelated PR.

References:
- [Apple TN3161: code-signing certificate selection](https://developer.apple.com/documentation/technotes/tn3161-inside-code-signing-certificates)
- [Apple SecKeychainItemCopyAccess implementation](https://github.com/apple-oss-distributions/Security/blob/main/OSX/libsecurity_keychain/lib/SecKeychainItem.cpp#L308)
- [Apple key ACL decoding](https://github.com/apple-oss-distributions/Security/blob/main/securityd/src/kckey.cpp#L190)
- [Apple Keychain decode/unlock path](https://github.com/apple-oss-distributions/Security/blob/main/securityd/src/kcdatabase.cpp#L1773)

Related to #76; intentionally does not close it.


mxcl:
One possible way forward is an explicit **Thorough Scan** button, separating checks that may prompt from the routine Scan.

- **Scan:** background-safe checks that do not prompt. The code-signing ACL check would stay out of this path while its no-prompt limitation remains unresolved.
- **Thorough Scan:** started explicitly by the user, with advance guidance: “Some checks may trigger permission or Keychain prompts. Choose Deny or Cancel. Don’t enter passwords or grant access.” These checks must never run automatically.

A denied or canceled prompt should produce **“Could not inspect”**, not “protected” or a clean result. Permission to inspect a key’s ACL and permission to use that key are different, so denial cannot establish that unattended signing is prevented.

Thorough Scan would still inspect metadata only: no signing, Secret retrieval, or access-control changes. It should offer a Stop button and handle blocked checks so an unanswered system prompt cannot leave the scan running indefinitely. We would need to verify what cancellation can actually stop, including any outstanding system prompt.

This is a proposed way to make the current detector useful without presenting it as safe for background Scans. The finding would remain **medium severity**. It does not resolve the legacy API’s prompt race; the PR remains a draft pending a concrete design and validation of this interaction.

